PR #30, opened 2015-02-04
Added support for linting HAML files

skateman:


app/workers/commit_monitor_handlers/commit_range/hamllint_checker.rb
```diff
@@ -0,0 +1,109 @@
...
+    files        = diff_details.keys
+
+    if files.length == 0
+      @results = {"files" => []}
+    else
+      @results = rubocop_results(files)
+      @results = RubocopResultsFilter.new(@results, diff_details).filtered
+    end
+
+    write_to_github
```
Fryguy: I think this checker should work slightly differently that the rubocop one in that it shouldn't write any comment at all if there were no haml files in the PR.  What do you think?

app/workers/commit_monitor_handlers/commit_range/hamllint_checker.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def write_to_github
+    logger.info("#{self.class.name}##{__method__} Updating pull request #{branch.pr_number} with rubocop comment.")
+
+    new_comments = MessageBuilder.new(results, branch).messages
+
+    branch.repo.with_github_service do |github|
+      @github = github
+      delete_github_comments(find_old_github_comments)
```
Fryguy: We don't want to kill the regular old rubocop comments, only the hamllint ones.

Fryguy:
@skateman This is awesome!  Really looking forward to this getting in.

skateman:
Maybe the idea to create a separate worker isn't the best. 
The *haml-lint* gem is just an extension of the *rubocop*, so we can just extend our [original checker](https://github.com/ManageIQ/miq_bot/blob/master/app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb#L61) to return with the *rubocop* and *haml-lint* results merged.

Fryguy:
Interesting...merging them together might be better, but yeah, we need the json support.

skateman:
@Fryguy can you test this PR with [this branch](https://github.com/causes/haml-lint/pull/63) of haml-lint?

Fryguy:
@skateman Sure.  We may want to make some specs as well.  Additionally I can give you full access to the sandbox repo

skateman:
@Fryguy it was merged into the upstream master!

Fryguy:
Great work!

jrafanie:
What else is needed here @skateman @Fryguy?

Fryguy:
@skateman Is this still WIP or is this ready to go?

skateman:
@Fryguy let's talk about this on IRC tomorrow

skateman:
@Fryguy @jrafanie I've updated the specs and I think it's ready.

jrafanie:
Yay, thanks @skateman.  I'll work with @Fryguy to get this merged.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -33,7 +33,16 @@ def process_branch
...
-      @results = rubocop_results(files)
+
+      @results = linter_results('rubocop', :files => files, :format => 'json')
+      haml = linter_results('haml-lint', :files => files, :reporter => 'json')
+
+      # Merge RuboCop and haml-lint results
+      %w(offense_count target_file_count inspected_file_count).each do |m|
+        @results['summary'][m] += haml['summary'][m]
+      end
+      @results['files'] += @haml['files']
```
Fryguy: Is this `@haml` right? It's a local variable so I think it sold not have the @.

Fryguy: Also if it's wrong then that means the specs didn't cost this case

spec/support/rubocop_spec_helper.rb
```diff
@@ -1,6 +1,15 @@
...
-  File.write(rubocop_json_file, `rubocop --format=json #{rubocop_check_path}`) unless File.exist?(rubocop_json_file)
+  unless File.exist?(rubocop_json_file)
+    rubocop = JSON.parse(`rubocop --format=json #{rubocop_check_path}`)
+    hamllint = JSON.parse(`haml-lint --reporter=json #{rubocop_check_path}`)
+
+    %w(offense_count target_file_count inspected_file_count).each do |m|
+      rubocop['summary'][m] += hamllint['summary'][m]
+    end
+    rubocop['files'] += hamllint['files']
+    File.write(rubocop_json_file, rubocop.to_json)
```
Fryguy: Wondering if somehow we can reuse the code in the model so we are not duplicating the code here,  though this may not be needed for this  PR

Gemfile.lock
```diff
@@ -200,13 +205,13 @@ GEM
       rspec-core (~> 2.14.0)
       rspec-expectations (~> 2.14.0)
       rspec-mocks (~> 2.14.0)
-    rubocop (0.27.1)
+    rubocop (0.30.0)
```
Fryguy: Can we avoid updating rubocop in this PR, or is this version a requirement?  Basically, just reset your Gemfile.lock, and then run bundle update haml-lint and it should update the minimal.

Fryguy:
Will this link to any documentation on the built message?

Fryguy:
This looks really good, and I think I can probably merge as is, except for the Gemfile.lock change.  I'm just not ready to incur any of the changes from a newer rubocop at this moment, since that usually means I need to update the .rubocop.yml files to filter new things out.

PR #36, opened 2015-02-25
Add TravisBuildKiller which kills dup builds for the same PR/branch.

### LINKED ISSUE #22: Travis: Find multiple jobs for the same PR and kill all old ones.

Fryguy:
Closes #22

### DIFF
--- a/app/workers/travis_build_killer.rb
+++ b/app/workers/travis_build_killer.rb
@@ -1,0 +1,35 @@
+class TravisBuildKiller
+  include Sidekiq::Worker
+  include Sidetiq::Schedulable
+  include MiqToolsServices::SidekiqWorkerMixin
+  sidekiq_options :queue => :miq_bot, :retry => false
+
+  recurrence { hourly.minute_of_hour(0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55) }
+
+  def perform
+    if !first_unique_worker?
+      logger.info "#{self.class} is already running, skipping"
+    else
+      kill_builds
+    end
+  end
+
+  private
+
+  def kill_builds
+    repo = CommitMonitorRepo.where(:upstream_user => "ManageIQ", :name => "manageiq").first
+    repo.with_travis_service do |travis|
+      builds_to_cancel = travis.builds
+        .take_while { |b| b.pending? || b.canceled? }
+        .reject(&:canceled?)
+        .group_by(&:pull_request_number)
+        .flat_map { |_pr_number, builds| builds[1..-1] }
+
+      builds_to_cancel.each do |b|
+        for_what = b.pull_request? ? "PR ##{b.pull_request_number}" : "merge commit"
+        logger.info "Canceling Travis build ##{b.number} for #{for_what}"
+        b.cancel
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ gem 'sinatra', require: false
 gem 'slim'
 
 gem 'miq_tools_services', :git => "git://github.com/ManageIQ/miq_tools_services.git", :branch => "master"
+gem 'travis'
 
 gem 'awesome_spawn'
 gem 'default_value_for'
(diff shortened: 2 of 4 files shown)


Fryguy:
hardcoded to ManageIQ/manageiq for now.  We could technically loop through all repos, but I'd rather rewrite this as reactive via the commit monitor, and be selective, but this is a first step.

app/workers/travis_build_killer.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def kill_builds
+    repo = CommitMonitorRepo.where(:upstream_user => "ManageIQ", :name => "manageiq").first
+    repo.with_travis_service do |travis|
+      builds_to_cancel = travis.builds
+        .take_while { |b| b.pending? || b.canceled? }
+        .reject(&:canceled?)
```
brandondunne: Why take the canceled builds then reject them?

app/workers/travis_build_killer.rb
```diff
@@ -0,0 +1,38 @@
...
+        .group_by(&:pull_request_number)
+        .flat_map { |_pr_number, builds| builds[1..-1] }
+
+      if builds_to_cancel.any?
+        builds_to_cancel.each do |b|
+          for_what = b.pull_request? ? "PR ##{b.pull_request_number}" : "merge commit"
+          logger.info "Canceling Travis build ##{b.number} for #{for_what}"
+        end
+
+        builds_to_cancel.each(&:cancel)
```
brandondunne: Sounds like this should be right below the ```logger.info``` line as ``` b.cancel```

Fryguy:
@brandondunne Fixed

brandondunne:
:+1: nice!

app/workers/travis_build_killer.rb
```diff
@@ -0,0 +1,35 @@
...
+  private
+
+  def kill_builds
+    repo = CommitMonitorRepo.where(:upstream_user => "ManageIQ", :name => "manageiq").first
+    repo.with_travis_service do |travis|
+      builds_to_cancel = travis.builds
+        .take_while { |b| b.pending? || b.canceled? }
+        .reject(&:canceled?)
+        .group_by(&:pull_request_number)
+        .flat_map { |_pr_number, builds| builds[1..-1] }
```
jrafanie: @Fryguy Should we sort by `number` before we choose all but the first build for the PR?

app/workers/travis_build_killer.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private
+
+  def kill_builds
+    repo = CommitMonitorRepo.where(:upstream_user => "ManageIQ", :name => "manageiq").first
+    repo.with_travis_service do |travis|
+      builds_to_cancel = travis.builds
+        .take_while { |b| b.pending? || b.canceled? }
+        .reject(&:canceled?)
```
jrafanie: @Fryguy I don't understand this, why are we taking pending and canceled builds and then rejecting canceled?

jrafanie:
This is awesome sauce.  :heart:

app/workers/travis_build_killer.rb
Fryguy: No, all the builds are already sorted.

app/workers/travis_build_killer.rb
brandondunne: @jrafanie I had the same question.  ```take_while``` will get you all of the pending or canceled builds until the last successful / errored / failed build (all of the ones waiting to be built) then throw away the ones that are already canceled.

Fryguy: Ok, so when we walk the builds it goes by build number order, If we stop at a canceled build, we'll miss a ton of pending builds.

screen shot 2015-02-25 at 5 44 19 pm

So in that screenshot, if we stop at a canceled build, we would only get the first 8 instead of everything up to the first green one.  Once we have everything, we throw away all the canceled and should be left with pending builds.

jrafanie: Thanks for the explanation!!! It makes sense now.

app/workers/travis_build_killer.rb
jrafanie: :tada: Yay

jrafanie:
:+1:

brandondunne:
Travis is broken, tests pass locally :+1:

PR #53, opened 2015-05-07
Add mail notification on labels

oschreib:
Add the option to alert users on a new issue added to requested label.

### DIFF
--- a/app/mailers/user_notifier.rb
+++ b/app/mailers/user_notifier.rb
@@ -1,0 +1,12 @@
+class UserNotifier < ActionMailer::Base
+  default :from => 'any_from_address@example.com'
+
+  # Send notification email about a new issue
+  def send_notification_email(email, issue, label)
+    @email = email
+    @issue = issue
+    @label = label
+    mail( :to => @email,
+          :subject => "New issue (#{issue}) labeled #{label} created" )
+  end
+end
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -3,3 +3,14 @@ require File.expand_path('../application', __FILE__)
 
 # Initialize the Rails application.
 MiqBot::Application.initialize!
+
+# Action
+ActionMailer::Base.smtp_settings = {
+    :user_name => 'username',
+    :password => 'password',
+    :domain => 'yourdomain.com',
+    :address => 'smtp.domain.com',
+    :port => 587,
+    :authentication => :plain,
+    :enable_starttls_auto => true
+}
--- a/config/labels.yaml
+++ b/config/labels.yaml
@@ -1,0 +1,8 @@
+label1:
+  mailto:
+    - ofer@somewhere.com
+    - foo@bar.com
+
+label2:
+  mailto:
+    - hapoel@katamon.com
--- a/lib/bot/issue_manager.rb
+++ b/lib/bot/issue_manager.rb
@@ -11,6 +11,7 @@ class IssueManager
   include GitHubApi
 
   ISSUE_MANAGER_YAML_FILE = File.join(File.dirname(__FILE__), 'config/issue_manager.yml')
+  LABELS_YAML_FILE = File.join(File.dirname(__FILE__), 'config/labels.yml')
   ORGANIZATION = "ManageIQ"
 
   COMMANDS = Hash.new do |h, k|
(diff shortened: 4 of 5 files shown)


oschreib:
Few questions: What's the recommended way to test this patch?
Also, How can I get the URL for the issue? (rather than it's id)

oschreib:
cc @simon3z

config/environment.rb
```diff
@@ -3,3 +3,14 @@
 
 # Initialize the Rails application.
 MiqBot::Application.initialize!
+
+# Action
+ActionMailer::Base.smtp_settings = {
+    :user_name => 'username',
```
brandondunne: Please use 2 space indentation and align the hash rockets

app/mailers/user_notifier.rb
```diff
@@ -0,0 +1,12 @@
+class UserNotifier < ActionMailer::Base
+  default :from => 'any_from_address@example.com'
```
brandondunne: We use Rails Config for the app, please fetch this value from there.

app/mailers/user_notifier.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  # Send notification email about a new issue
+  def send_notification_email(email, issue, label)
+    @email = email
+    @issue = issue
+    @label = label
+    mail( :to => @email,
+          :subject => "New issue (#{issue}) labeled #{label} created" )
```
brandondunne: This subject seems very specific.  We will probably want to add other notification e-mails in the future, should this be in a subclass or something?

config/environment.rb
```diff
@@ -3,3 +3,14 @@
 
 # Initialize the Rails application.
 MiqBot::Application.initialize!
+
+# Action
+ActionMailer::Base.smtp_settings = {
```
brandondunne: This belongs in an Initializer, not environment.rb

config/environment.rb
```diff
@@ -3,3 +3,14 @@
...
+    :user_name => 'username',
+    :password => 'password',
```
brandondunne: Again, please fetch values from Rails Config

lib/bot/issue_manager.rb
```diff
@@ -11,6 +11,7 @@ class IssueManager
   include GitHubApi
 
   ISSUE_MANAGER_YAML_FILE = File.join(File.dirname(__FILE__), 'config/issue_manager.yml')
+  LABELS_YAML_FILE = File.join(File.dirname(__FILE__), 'config/labels.yml')
```
brandondunne: Again, Rails Config

app/mailers/user_notifier.rb
brandondunne: To elaborate on Rails Config...

There is a config/settings.yml file that is a blank template.  In production, we deploy a config/settings.local.yml that overrides the values.  See [here](https://github.com/ManageIQ/miq_bot/blob/c524fbf51ccffb32a464bd0cdcb5a797cdf80eb9/app/models/commit_monitor_repo.rb#L49) for an example.

lib/bot/issue_manager.rb
```diff
@@ -236,12 +240,26 @@ def get_credentials
...
+            UserNotifier.send_notification_email(address, issue.number, label).deliver
+          end
+        end
+      end
+    rescue Exception => msg
+      logger.error("Couldn't send email notification. Exception: #{msg}")
+    end
+  end
+
+  def load_yaml_file(file)
```
brandondunne: This change shouldn't be needed anymore

lib/bot/issue_manager.rb
```diff
@@ -236,12 +240,26 @@ def get_credentials
...
     begin
-      @timestamps = YAML.load_file(ISSUE_MANAGER_YAML_FILE)
+      if @notify.has_key?(label)
+        if @notify[label].has_key?('mailto')
+          @notify[label]['mailto'].each do |address|
+            UserNotifier.send_notification_email(address, issue.number, label).deliver
+          end
+        end
+      end
+    rescue Exception => msg
```
brandondunne: Please rescue your specific exception rather than the generic "Exception"

config/labels.yaml
```diff
@@ -0,0 +1,8 @@
+label1:
```
brandondunne: Shouldn't need this yaml file if we use Rails Config

lib/bot/issue_manager.rb
oschreib: Not needed? because of the rails config? 
I guess we still need to call the send_notification_email

config/environment.rb
oschreib: Any favorite location for that initializer?

brandondunne: Initializers live in config/initializers/ you can read more [here](http://guides.rubyonrails.org/configuring.html#using-initializer-files)

brandondunne:
@oschreib Cool idea, however this will only send an e-mail when labels are added using the "@miq_bot add_label" comment.  In the future, it would be nice to leverage github notifications to send an e-mail rather than doing it through rails.

brandondunne:
@Fryguy should be able to help you set up the environment for testing

oschreib:
Had another idea how to implement this without adding the smtp config, stay tuned.

PR #54, opened 2015-05-09
Enhance gemfile checker

### LINKED ISSUE #3: When detecting gem changes, set the gem_changes label.

jrafanie:
* Configure the repository that GemfileChecker monitors.
  * This makes it easier to test "ManageIQ/sandbox" as an example.
* Delete prior Gemfile comments before adding a new one.
* Add the "gem changes" label to the PR if it's not already applied.
  * Fixes #3

cc @Fryguy 

Once this in, we can extract much of the code that is similar to the rubucop checker that detects the already existing comments.

### DIFF
--- a/app/workers/commit_monitor_handlers/commit/gemfile_checker.rb
+++ b/app/workers/commit_monitor_handlers/commit/gemfile_checker.rb
@@ -6,7 +6,7 @@ class CommitMonitorHandlers::Commit::GemfileChecker
     [:pr]
   end
 
-  attr_reader :branch, :commit
+  attr_reader :branch, :commit, :github, :pr
 
   def perform(branch_id, commit, commit_details)
     @branch = CommitMonitorBranch.find(branch_id)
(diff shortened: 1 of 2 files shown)


app/workers/commit_monitor_handlers/commit/gemfile_checker.rb
```diff
@@ -32,21 +36,65 @@ def tag
...
+      add_pr_label
     end
   end
 
+  def github_org_repo
+    [branch.repo.upstream_user, branch.repo.name]
+  end
+
+  def add_pr_label
+    issue_labels = github.issues.labels.list(*github_org_repo, "issue_id" => pr)
```
Fryguy: The branch should be associated with a PR, so you can get the issue first, and then get the labels and not need the repo info I think...Plus the github handle comes from the branch object with *should* already have this info...

Something like

```ruby
issue = github.issues("issue_id" => pr)
issue.labels.list
```

config/settings.yml
```diff
@@ -9,5 +9,6 @@ commit_monitor:
   bugzilla_product:
 gemfile_checker:
   pr_contacts: []
+  enabled_repos: ["ManageIQ/manageiq"]
```
Fryguy: Can you just leave this empty to being with, and we can change in deployment.

jrafanie:
@Fryguy see last commit, reworked it from your suggestions.

* Get the issue itself and ask it what labels it has.
* Drop attr_reader on LABEL_NAME in favor of using the constant instead.

* I also defaulted the gemfile_checker enabled_repos to [] as suggested.

Finally, we can "skip" the "is this label already applied?" check and just blindly add the label but the github interface looks kinda clunky since it will say the bot added the label twice in a row.  What do you think?

See below:

screen shot 2015-05-12 at 3 30 45 pm

Fryguy:
Oh well, that's ugly.  Hmm...then maybe we should have the check :disappointed:

Fryguy:
The only other option is [replacing all labels](https://developer.github.com/v3/issues/labels/#replace-all-labels-for-an-issue), but you'd still have to look them all up, so I don't think you gain anything by that.

jrafanie:
Yeah, I'm fine with the current behavior of looking up the label before adding it.  We can always change it if need be when we do other "more frequent" labeling.

PR #55, opened 2015-05-10
Add labels subscription

oschreib:
Add the option to configure subscribers to specific labels. Once
subscribed to a label, miq_bot will add them to the cc list on every new
PR with configured label.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -11,3 +11,5 @@ gemfile_checker:
   pr_contacts: []
 issue_manager:
   repo_names: []
+label_notifier:
+  contacts: { label1: [] }
--- a/lib/bot/issue_manager.rb
+++ b/lib/bot/issue_manager.rb
@@ -144,6 +144,20 @@ EOMSG
     end
     if !valid_labels.empty?
       issue.add_labels(valid_labels)
+      notify_on_label(issue, valid_labels)
+    end
+  end
+
+  def notify_on_label(issue, labels)
+    labels.each do |label|
+      if Settings.label_notifier.contacts[label] &&
+         Settings.label_notifier.contacts[label] > 0
+        message = 'cc '
+        Settings.label_notifier.contacts[label].each do |contact|
+          message += "@#{contact}"
+        end
+        issue.add_comment(message)
+      end
     end
   end
 


oschreib:
cc @simon3z @abonas

lib/bot/issue_manager.rb
```diff
@@ -144,6 +144,19 @@ def add_labels(command_value, author, issue)
     end
     if !valid_labels.empty?
       issue.add_labels(valid_labels)
+      notify_on_label(issue, valid_labels)
+    end
+  end
+
+  def notify_on_label(issue, labels)
```
simon3z: @oschreib  does this work only for issues or for pull requests as well? (We want it for both)

config/settings.yml
```diff
@@ -11,3 +11,5 @@ gemfile_checker:
   pr_contacts: []
 issue_manager:
   repo_names: []
+label_notifier:
+  labelname: []
```
abonas: 1. please change to have an underscore between 2 words (ruby style)
2. switch to plural? (label_names)

lib/bot/issue_manager.rb
```diff
@@ -144,6 +144,19 @@ def add_labels(command_value, author, issue)
...
+    labels.each do |label|
+      if Settings.label_notifier[label]
+        message = "cc "
+        Settings.label_notifier[label].each do |contact|
```
abonas: does this consider a case where contacts list is empty?

abonas: also, an alternative syntax:
(where contacts is Settings.label_notifier[label])
msg = "cc " << contacts.map{ |contact| "@#{contact}"}.join(" ")

lib/bot/issue_manager.rb
oschreib: Seems that yes, I didn't see any other entry point for labels/

config/settings.yml
oschreib: well, the idea is to have multiple <labelname>: [contactlist]., so this is just an example.

lib/bot/issue_manager.rb
oschreib: There's no point in having an empty contact list (just remove the label).
about the second syntax - I know it, but to me it's really hard to read.

oschreib:
cc @Fryguy

simon3z:
@Fryguy @brandondunne can you review/merge?

config/settings.yml
```diff
@@ -11,3 +11,5 @@ gemfile_checker:
   pr_contacts: []
 issue_manager:
   repo_names: []
+label_notifier:
+  contacts: { label1: [] }
```
Fryguy: Please just leave this as an empty value.  `contacts: {}`

The settings are actually used, so the bot will just start looking for a label called "label1"

lib/bot/issue_manager.rb
```diff
@@ -144,6 +144,20 @@ def add_labels(command_value, author, issue)
...
+    end
+  end
+
+  def notify_on_label(issue, labels)
+    labels.each do |label|
+      if Settings.label_notifier.contacts[label] &&
+         Settings.label_notifier.contacts[label] > 0
+        message = 'cc '
+        Settings.label_notifier.contacts[label].each do |contact|
+          message += "@#{contact}"
```
Fryguy: This can be simplified to

```ruby
users = Settings.label_notifier.contacts[label].map {|c| "@#{c}"}.join(" ")
message = "cc #{users}"
```

lib/bot/issue_manager.rb
```diff
@@ -144,6 +144,20 @@ def add_labels(command_value, author, issue)
...
+        end
+        issue.add_comment(message)
```
Fryguy: This is going to create a separate comment per label in addition to the initial comment response.  Can we somehow tack it into the regular response so we don't flood the issue?

lib/bot/issue_manager.rb
```diff
@@ -144,6 +144,20 @@ def add_labels(command_value, author, issue)
...
     if !valid_labels.empty?
       issue.add_labels(valid_labels)
+      notify_on_label(issue, valid_labels)
+    end
+  end
+
+  def notify_on_label(issue, labels)
+    labels.each do |label|
+      if Settings.label_notifier.contacts[label] &&
+         Settings.label_notifier.contacts[label] > 0
```
brandondunne: I think this could be simplified to
```ruby
if Settings.label_notifier.contacts[label].kind_of?(Array)
```

pilhuhn:
@Fryguy Would that be merged if someone would fix the comments above?

bdunne:
Closing due to inactivity, please reopen with the suggested changes if this feature is still desired.

PR #59, opened 2015-05-19
Restart stalled travis builds automagically - via travis event monitoring

jrafanie:
Travis has a pretty annoying recurring issue that results in the following "stalled/timed out" build message:

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

To be clear, we know these errors several times per week but we don't have an exact number because they happen to everyone randomly.  They occur on both branch and pull request builds.  After playing with the travis bus, I see that we are clearly not the only ones with this problem as random non-ruby projects also get these errors.

Unfortunately, it's very difficult to diagnose as it sometimes occurs outside of our tests, such as during installation of gems or setting up the rvm environment in travis.  Without any means to easily detect this, developers have to go to the travis job that errored, look at the log, be surprised by the weird error message, contemplate what their PR did that caused it, ask others for help, and eventually ask others to restart the failed job.

This pull request hooks into the travis event bus, watching our repositories for all events.  These events are thrown on sidekiq for any handlers to react to them.  The StalledFinishedJob handler only looks at errored job events for open pull request, by hooking into our existing Commit Monitor and Pull Request Monitor, looking at the job logs for the specific string indicating a "stalled" job. If a stalled job is found, the bot will restart the job, log a message and write a github comment indicating it detected and restarted the stalled job.  This is also very fast.  Many times the the job is restarted before you can read the "stalled" log message in the travis UI.

Once this is in place, developers won't have to waste time visiting travis for stalled builds, get them restarted, and waiting for it to complete.  Additionally, we'll get an indication of how many "stalled" jobs we get based on the github comments.

To do items:
- [x] I need to test issuing the restart of the job with the travis auth setup - has been done with a standalone script.
- [x] I need to integrate with the commit monitor via the CommitMonitorBranch.  Need to check if the PR is still open before the stalled job checker tries to fetch and parse the job log.
- [x] I need to test the issue manager with the faraday and octokit gem upgrades
- [x] change 'travis_event' sidekiq queue back to use the 'miq_bot' one???
- [x] Need to serialize the raw event as a hash so workers can take whatever they want.

Thoughts?


lib/travis_event/handlers.rb
```diff
@@ -0,0 +1,12 @@
+require Rails.root.join('lib/faraday_patch')
```
Fryguy: Maybe we should just do this in an intitializer and make it simple (so we don't need lib on the autoload paths and stuff)

config/application.rb
```diff
@@ -20,5 +20,6 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
+    config.autoload_paths += ['lib']
```
jrafanie: @Fryguy This is needed so we can constantize the handlers [here](https://github.com/ManageIQ/miq_bot/pull/59/files#diff-be7efa9d2bb56731b17b2413e503b575R6)

lib/travis_event/handlers/stalled_finished_job.rb
```diff
@@ -0,0 +1,116 @@
...
+        # Local checks: skip unknown repos or non-current branches
+        return unless current_commit_monitor_branch(raw_branch_name)
+
+        # Remote checks: Skip missing travis repo or job and non-stalled builds
+        return unless travis_repo = Travis::Repository.find(repo_name)
+        return unless @job = find_job(travis_repo, number)
+        return unless stalled?
+
+        message = "Restarting stalled job: [#{repo_name}##{number}]"
+        commit_monitor_branch.write_github_comment(message)
```
jrafanie: @Fryguy notice, I added a placeholder for commenting on a PR CommitMonitorBranch. It seemed like the place to do it.

app/models/commit_monitor_branch.rb
```diff
@@ -31,4 +31,11 @@ def last_commit_uri
   def pr_number
     MiqToolsServices::MiniGit.pr_number(name) if pull_request?
   end
+
+  def write_github_comment(message)
+    return unless pull_request?
+
+    logger.info("#{self.class.name}##{__method__} [#{job.inspect_info}] would write comment for PR: #{pr}")
+    # TODO: Write the comment
```
jrafanie: @Fryguy this will need to go through the message building logic eventually

jrafanie:
@Fryguy I added basic linkage between the travis event processing and the CommitMonitorBranch information.  The first cut of this can just comment on the PR blindly (without deleting the old comment) or can use the batch message support when that lands.

jrafanie:
@Fryguy If it looks ok, I can clean up some of the rubocop comments and implement a basic github commenter much like the GemfileChecker does we can later refactor into the batch message code.  I was probably going to test the faraday upgrade with the issue manager and test these changes on the PR and be done with it for now.  Thoughts?

Fryguy:
I extracted the message building logic in #60

jrafanie:
@Fryguy got it working end to end (comment + restart job) on sandbox locally.  Below is a screenshot from https://github.com/ManageIQ/sandbox/pull/145 which simulates a stalled build by sleeping for 15 minutes in a test.

screen shot 2015-06-02 at 11 48 44 pm

Gemfile
```diff
@@ -36,7 +36,7 @@ gem 'sinatra', require: false
 gem 'slim'
 
 gem 'miq_tools_services', :git => "git://github.com/ManageIQ/miq_tools_services.git", :branch => "master"
-gem 'travis'
+gem 'travis', '~>1.7.6'
```
jrafanie: fixes 406 error

Gemfile
```diff
@@ -51,9 +51,9 @@ group :development, :test do
 end
 
 group :issue_manager do
-  gem 'octokit',       '~> 1.25.0'
+  gem 'octokit',       '~> 3.8.0'
   gem 'minigit',       '~> 0.0.4'
 
   # Lock down dependency
-  gem 'faraday', '~> 0.8.8'
+  gem 'faraday', '~> 0.9.1'
```
jrafanie: required by newer travis

Gemfile
```diff
@@ -51,9 +51,9 @@ group :development, :test do
 end
 
 group :issue_manager do
-  gem 'octokit',       '~> 1.25.0'
+  gem 'octokit',       '~> 3.8.0'
```
jrafanie: old octokit uses old faraday api, so, we need to upgrade this to use the newer faraday api

Procfile
```diff
@@ -1,4 +1,5 @@
 thin:                      bundle exec rails s -p $PORT
-sidekiq:                   bundle exec sidekiq -q miq_bot
+sidekiq:                   bundle exec sidekiq -q miq_bot, -q travis_event -L log/sidekiq.log
```
jrafanie: added travis_event queue

Procfile
```diff
@@ -1,4 +1,5 @@
...
 issue_manager:             bundle exec ruby lib/bot/miq_bot.rb
+travis_listener:           bundle exec rails runner lib/travis_event/listener.rb
```
jrafanie: launch the separate travis event listener process that waits for events from travis for the subscribed repos and sends any events to any event handlers to decide if they can process it.

lib/bot/githubapi/git_hub_api.rb
```diff
@@ -20,11 +20,11 @@ def self.connect(username, password)
   end
 
   def self.execute(client, cmd, *args)
-    rate_limit_remaining = client.rate_limit_remaining
+    rate_limit_remaining = client.rate_limit.remaining
```
jrafanie: octokit api deprecated the old syntax.

Procfile
Fryguy: yeah we should remove this.

Fryguy: The log line has to go too or it will break in production.

app/models/commit_monitor_branch.rb
```diff
@@ -31,4 +31,19 @@ def last_commit_uri
   def pr_number
     MiqToolsServices::MiniGit.pr_number(name) if pull_request?
   end
+
+  def write_github_comment(message, header, continuation_header = nil, replace = true)
+    return unless pull_request?
```
Fryguy: This should probably raise an exception if it's not a PR.

app/models/commit_monitor_branch.rb
```diff
@@ -31,4 +31,19 @@ def last_commit_uri
...
+
+    message_builder = MiqToolsServices::Github::MessageBuilder.new(header + message, continuation_header)
```
Fryguy: This is weird.  It should be:

```ruby
message_builder = MiqToolsServices::Github::MessageBuilder.new(header, continuation_header)
message_builder.write(message)
```

app/models/commit_monitor_branch.rb
```diff
@@ -31,4 +31,19 @@ def last_commit_uri
...
+
+    repo.with_github_service do |github|
+      @github = github
```
Fryguy: No need for this if you're not extracting to methods.

app/models/commit_monitor_branch.rb
```diff
@@ -31,4 +31,19 @@ def last_commit_uri
...
+      # TODO: Refactor the common "delete prior tagged issues" into miq_tools_services
+      # github.delete_issue_comments(pr_number, header) if replace
+      github.create_issue_comments(pr_number, message_builder.comments)
+    end
+
+    logger.info("#{self.class.name}##{__method__} Wrote comment with header: #{header} on PR: #{name}")
```
Fryguy: Would be better to have the logging before the execution.  That way if there's a failure we can trace back to what it was trying to do.

lib/travis_event/handlers/stalled_finished_job.rb
```diff
@@ -0,0 +1,114 @@
+require 'travis/client'
+
+module TravisEvent
+  module Handlers
+    class StalledFinishedJob
```
Fryguy: Sidekiq workers should live in the workers directory...that's where they are expected to live

lib/travis_event/handlers/stalled_finished_job.rb
```diff
@@ -0,0 +1,114 @@
...
+      include Sidekiq::Worker
+      sidekiq_options :queue => :travis_event, :retry => 5
+
+      STALLED_BUILD_TEXT = "\n\nNo output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.\n\nThe build has been terminated\n\n"
```
Fryguy: Do we want to find a smaller substring?

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
+module TravisEvent
+  module Handlers
+    def self.handlers_for
```
Fryguy: Why not just `.handlers`

Fryguy: In fact, why have a separate module at all for one method.

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+  class Listener
+    ALL_EVENTS = %w(build:created build:started build:finished job:created job:started job:finished)
+
+    # repos: [Travis::Repository.find("rails/rails")]
+    def self.monitor(*repos)
+      Travis.listen(*repos) do |stream|
+        stream.on(*ALL_EVENTS) do |event|
+          args = extract_args(event)
+          if args
+            Handlers.handlers_for.each do |handler|
```
Fryguy: You should cache the handlers in the class method like we did with the commit_monitor, or this will get expensive.

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+              handler.perform_async(*args)
+            end
+          end
+        end
+      end
+    end
+
+    def self.extract_args(event)
+      # Event = Struct.new(:type, :repository, :build, :job, :payload)
```
Fryguy: Is this comment needed?

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+          [event.build.number, event.build.state, event.build]
+        when 'job'
+          [event.job.number, event.job.state, event.job.build]
+        else
+          Sidekiq::Logging.logger.info("#{name}##{__method__} Discarding unsupported event source type: #{event.type}")
+          return nil
+        end
+
+      branch_name = build.pull_request? ? build.pull_request_number : build.commit.branch
+      [event.repository.slug, number, event.type, state, branch_name, build.pull_request?]
```
Fryguy: Can we just serialize the whole event somehow and simplify the params?  Even a simple nested Hash might be easier.

Fryguy: Instead of all this try:

```ruby
hash = event.to_h
hash.keys.each do |k|
  v = h.delete(k)
  v = v.attributes if v.respond_to?(:attributes)
  h[k.to_s] = v
end
```

Fryguy: The other option is to call .to_h instead of .attributes there and make all the Travis calls on the server side.  This can be useful if you end up with multiple workers, since it doesn't make sense for them *all* to lookup the build object for the same information.

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+end
+
+require 'travis'
+repos = Settings.travis_event.monitored_repos.collect do|repo|
+  Travis::Repository.find(repo)
+end
+
+TravisEvent::Listener.monitor(*repos)
```
Fryguy: I think the listener launch script should be separate from the listener class.  I think it's ok to live in lib, but the launcher should be in bin or lib/bin or something.

Fryguy: Also I fell like TravisEvent::Listener.monitor should just take repo names versus repo objects.

app/models/commit_monitor_branch.rb
jrafanie: This method was rewritten as the caller should pass in the header as the tag + first line.  The new code will look similar to your recommendation.

jrafanie:
@Fryguy I think this is finally ready...  The issue we had with restarting was because I was logged into travis as myself when I synced the miq-test org data.  It works after I logged into travis as the bot and synced the data.  Additionally, I simplified the with_travis_service, see comment above.

On Monday, we can test this with your changes on top for the other pull requests.

app/models/commit_monitor_repo.rb
```diff
@@ -45,9 +55,8 @@ def with_github_service
 
   def with_travis_service
     raise "no block given" unless block_given?
-    client = Travis::Client.new
-    client.github_auth(Settings.github_credentials.password) # Assumes password is holding a token
-    repo = client.repo(fq_name)
-    yield repo
+
+    Travis.github_auth(Settings.github_credentials.password)
+    yield Travis::Repository.find(fq_name)
```
jrafanie: Simplified this.  Travis::Repository.find is what we are returning anyway so call it directly instead of separately instantiating the client.  Note, on the repo, you can still get to the initialized client.

app/workers/travis_event_handlers/stalled_finished_job.rb
```diff
@@ -0,0 +1,95 @@
...
+    COMMENT_TAG = "<stalled_finished_job />".freeze
+
+    attr_reader :repo, :branch, :job, :slug, :number, :event_type, :state
+
+    def perform(event_hash)
+      @slug       = event_hash.fetch_path("payload", "repository_slug")
+      @number     = event_hash.fetch_path("payload", "number")
+      @event_type = event_hash.fetch_path("type")
+      @state      = event_hash.fetch_path("payload", "state")
+      pr_number   = event_hash.fetch_path("build", "pull_request_number")
```
jrafanie: Implemented the event_hash serialization on the receiver side.

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+        path.chomp(".rb").classify.constantize
+      end
+    end
+  end
+
+  class Listener
+    ALL_EVENTS = %w(build:created build:started build:finished job:created job:started job:finished)
+
+    # repos: ["rails/rails", "manageiq/manageiq"]
+    def self.monitor(*slugs)
```
jrafanie: Now accepts an array arg or several repos in the format of the "slug" (not the Travis::Repository instances).

lib/travis_event/listener.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+      end
+    end
+
+    def self.travis_repos(slugs)
+      slugs = slugs.first if slugs.first.kind_of?(Array)
+      slugs.collect { |slug| Travis::Repository.find(slug) }
+    end
+
+    def self.event_hash(event)
```
jrafanie: Implemented the serialization of the event into a generic event hash.

PR #67, opened 2015-06-11
Force haml-lint to use only RuboCop as his linter

skateman:


martinpovolny:
solves issues seen in https://github.com/ManageIQ/manageiq/pull/2075

Fryguy:
How does this solve those issues if only specs are  changed here.

skateman:
This removes the extensive misleading rubocop warnings from haml-lint as you can see in Martin's PR above.

skateman:
This can be also solved by changing the configuration file, so please do not merge!

PR #70, opened 2015-06-18
Updated haml-lint to 0.13.0

skateman:


Gemfile
```diff
@@ -42,7 +42,7 @@ gem 'awesome_spawn'
 gem 'default_value_for'
 gem 'more_core_extensions', :require => 'more_core_extensions/all'
 gem 'rubocop', '>= 0.27.1'
-gem 'haml-lint', '~> 0.12.0'
+gem 'haml-lint', '~> 0.13.0'
```
chessbyte: @Fryguy @skateman should rubocop and haml-lint be excluded from production?

Fryguy: What do you mean @chessbyte ?  We don't do rubocops on production yet as is.

chessbyte: my bad @Fryguy - I thought this was the manageiq repo.

PR #79, opened 2015-06-23
Add developer setup

imtayadeway:
This adds the steps required to get up and running in development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,106 @@ problems.  Some of the things it does are:
   - Ability to request tasks of the bot, such as setting of labels, assignment,
     and milestone, which would otherwise not be possible without commit rights
     to the repo.
+
+## Development
+
+### Prerequisites
+
+* ruby 1.9.3 with bundler
+* redis ~> 2.8.10
+* postgresql
+
+### Steps
+
+1. Create/fork a sandbox repository (replace `miq-test/sandbox` with your test
+   repository below):
+   ```
+   https://github.com/miq-test/sandbox
+   ```
+
+2. Fork https://github.com/ManageIQ/miq_bot
+
+3. Clone the miq_bot repo and add the upstream remote:
+    ```
+    git clone git@github.com:<your github handle>/miq_bot.git
+    cd miq_bot
+    git remote add upstream git@github.com:ManageIQ/miq_bot.git
+    git fetch —all
+    ```
+
+4. Add the sandbox app to the `repos` directory so it can be monitored:
+   ```
+   mkdir repos
+   git clone -o upstream git@github.com:miq-test/sandbox.git repos/sandbox
+   ```
+
+5. Install any gem dependencies:
+   ```
+   bundle install
+   ```
+
+6. Open up the `config/database.yml` file in your editor. Add the
+   `username` and `password` for your Postgresql database (If you set
+   up the ManageIQ app locally just use the same credentials).
+
+7. Set up the databases:
+   ```
+   bundle exec rake db:create:all
+   bundle exec rake db:migrate
+   RAILS_ENV=test bundle exec rake db:migrate
+   RAILS_ENV=production bundle exec rake db:migrate
+   ```
+
+8. Add the sandbox app to the database:
+   ```
+   bundle exec rails runner 'CommitMonitorRepo.create_from_github!("miq-test", "sandbox", Rails.root.join("repos", "sandbox"))'
+   ```
+
+9. Either start redis as a daemon or have foreman start it for you by
+   adding this to the `Procfile`:
+   ```yaml
+   # Procfile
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -312,3 +312,6 @@ DEPENDENCIES
   travis (~> 1.7.6)
   turbolinks
   uglifier (>= 1.3.0)
+
+BUNDLED WITH
+   1.10.4
```
jrafanie: @imtayadeway Let's keep this change out of this documentation only PR for now.

Procfile
```diff
@@ -1,5 +1,5 @@
 thin:                      bundle exec rails s -p $PORT
-sidekiq:                   bundle exec sidekiq -q miq_bot
+sidekiq:                   bundle exec sidekiq -L log/sidekiq.log -q miq_bot
```
jrafanie: Even though I agree this makes development easier, I believe @Fryguy or @brandondunne said this messes up the deployment mechanism we use on the actual bot.  @Fryguy @brandondunne Can we use an environment variable instead of a hardcoded path like we do the with thin port?

README.md
```diff
@@ -19,3 +19,102 @@ problems.  Some of the things it does are:
...
     to the repo.
+
+## Development
+
+### Prerequisites
+
+* Membership in the miq-test organization
+* ruby 1.9.3 with bundler
+* redis ~> 2.8.10
+* postgresql
```
jrafanie: Note: miq-test/sandbox is the example used as the github repo to monitor with the bot.

README.md
```diff
@@ -19,3 +19,102 @@ problems.  Some of the things it does are:
...
+
+3. Clone the miq_bot repo and add the upstream remote:
+    ```
+    git clone git@github.com:<your github handle>/miq_bot.git
+    cd miq_bot
+    git remote add upstream git@github.com:ManageIQ/miq_bot.git
+    git fetch —all
+    ```
+
+4. Add the sandbox app to the `repos` directory so it can be monitored:
```
jrafanie: Let's add `mkdir repos` here

README.md
```diff
@@ -19,3 +19,102 @@ problems.  Some of the things it does are:
...
+    # gemfile_checker
+    pr_contacts: [“@your_username"]
+    enabled_repos: ["miq-test/sandbox"]
+
+    # issue_manager
+    repo_names: ["sandbox"]
+
+    # travis_event
+    enabled_repos: ["miq-test/sandbox"]
+    ```
```
jrafanie: It's clear that this is messy... we will need a clearer way to specify the repo to monitor...
Maybe the CommitMonitorRepo model could eventually have a table of the enabled handlers.  This comment is fine for now

README.md
```diff
@@ -19,3 +19,102 @@ problems.  Some of the things it does are:
...
+    Then set to the following values:
+    ```yaml
+    # config/settings.local.yml:
+
+    # github_credentials
+    username: "miq-bot"
+    password: # token goes here
+
+    # gemfile_checker
+    pr_contacts: [“@your_username"]
```
jrafanie: The first set of double quotes is utf8 double quotes or something.

Procfile
brandondunne: Do we need to change the log file?

jrafanie: It's useful to specify it when you're developing locally so you don't have to manually tee or redirect the output to a log file.

README.md
imtayadeway: not sure why i presented it like this....could be better (at least for now) if i just reproduce my whole `config/settings.local.yml` file here as is, with the token obscured?

README.md
imtayadeway: shall i remove the miq-test org requirement, and make it more generic, i.e. something like you need _a_ sandbox repo to test it out?

imtayadeway:
@jrafanie i reverted the non-documentation commits and tweaked the README - can you rereview?

README.md
jrafanie: It's ok as is.  Just need to clean it up so it's easier to configure in the future.

README.md
jrafanie: I think it's fine to say something like "replace miq-test/sandbox" with your test repository below.

README.md
imtayadeway: :+1: done

imtayadeway:
@jrafanie any other comments? I'll squash this down if you think it's OK

jrafanie:
Looks great @imtayadeway, please squash the commits and update the PR description.  Thanks!

:+1:

imtayadeway:
should be GTG, @jrafanie

jrafanie:
Ah, @imtayadeway use [skip ci] in your commit message next time so travis doesn't need to run.

imtayadeway:
@jrafanie i saw this after it was too late. :cry: forgot that i had removed everything but documentation!

PR #80, opened 2015-06-24
Make gemfile checker a range handler.

### LINKED ISSUE #78: Gemfile checker should be a commit_range handler, not a commit handler
If you have two commits each with a new gem added to the Gemfile, gemfile checker will comment on the PR twice instead of once.

See below
screen shot 2015-06-23 at 2 59 33 pm

This came from this PR:  https://github.com/miq-test/sandbox/pull/3/

carbonin:
The gemfile checker was a commit handler, so could generate more noise than was
required for PRs that had Gemfile changes over several commits. This changes it
to a commit range handler which allows only one comment for multiple Gemfile
changes.

https://trello.com/c/ts3G5COh/11-miq-bot-gemfile-checker-should-be-a-commit-range-handler-not-commit-handler
Fixes #78 

Below screenshot shows it now doing a single comment with a range of commits:

screen shot 2015-06-26 at 2 43 12 pm

### DIFF
--- a/app/models/commit_monitor_branch.rb
+++ b/app/models/commit_monitor_branch.rb
@@ -11,6 +11,8 @@ class CommitMonitorBranch < ActiveRecord::Base
   default_value_for(:commits_list) { [] }
   default_value_for :mergeable, true
 
+  delegate :enabled_for?, :to => :repo
+
   def self.with_branch_or_pr_number(n)
     n = MiqToolsServices::MiniGit.pr_branch(n) if n.kind_of?(Fixnum)
     where(:name => n)
@@ -18,11 +20,6 @@ class CommitMonitorBranch < ActiveRecord::Base
 
   def self.github_commit_uri(user, repo, sha = "$commit")
     "https://github.com/#{user}/#{repo}/commit/#{sha}"
-  end
-
-  def enabled_for?(checker)
-    repos = Settings.public_send(checker).enabled_repos
-    repo.fq_name.in?(repos)
   end
 
   def last_commit=(val)
--- a/app/models/commit_monitor_branch.rb
+++ b/app/models/commit_monitor_branch.rb
@@ -20,6 +20,11 @@ class CommitMonitorBranch < ActiveRecord::Base
     "https://github.com/#{user}/#{repo}/commit/#{sha}"
   end
 
+  def enabled_for?(checker)
+    repos = Settings.public_send(checker).enabled_repos
+    repo.fq_name.in?(repos)
+  end
+
   def last_commit=(val)
     super
     self.last_changed_on = Time.now.utc if last_commit_changed?
--- a/app/workers/commit_monitor_handlers/commit_range/gemfile_checker.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/gemfile_checker.rb
@@ -1,4 +1,4 @@
-class CommitMonitorHandlers::Commit::GemfileChecker
+class CommitMonitorHandlers::CommitRange::GemfileChecker
   include Sidekiq::Worker
   sidekiq_options :queue => :miq_bot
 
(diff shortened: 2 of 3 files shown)


jrafanie:
For future reference, you can automatically close an issue when your PR is merged by using the "Fixes #78" syntax in the commit message.  See: https://help.github.com/articles/closing-issues-via-commit-messages/

jrafanie:
Please fix up the whitespace rubocop issues.

app/workers/commit_monitor_handlers/commit_range/gemfile_checker.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        files = diff_details_for_branch.keys
+        return unless files.any? { |f| File.basename(f) == "Gemfile" }
+
+        process_branch
+      end
+
+      private
+
+      def repo_enabled?
```
jrafanie: This seems like something we might want to extract to a common place as we add more an more checkers that need to "configure which repos they want to check".

app/workers/commit_monitor_handlers/commit_range/gemfile_checker.rb
```diff
@@ -0,0 +1,101 @@
...
+        [:pr]
+      end
+
+      attr_reader :branch, :commits, :github, :pr
+
+      def perform(branch_id, _new_commits)
+        @branch = CommitMonitorBranch.where(:id => branch_id).first
+        
+        if @branch.nil?
+          logger.info("Branch #{branch_id} no longer exists.  Skipping.")
```
jrafanie: Can you review and normalize the log message prefix for the checker?  This wasn't clear before.   I believe sidekiq will prefix all of your messages in the worker with the `GemfileChecker` value... can you grep your logs to see if it's consistent throughout this worker?  I'm not sure if the `self.class` used below in the log lines will be redundant.

jrafanie:
Looks really good @carbonin @imtayadeway, minor nitpicks above.  :+1:

imtayadeway:
@jrafanie we took care of most of the issues - just the AbcSize cop that we left as that will take more time and care to refactor that probably all the checkers could benefit from. Any further issues?

Fryguy:
I should just turn off those complexity checkers :)

app/models/commit_monitor_branch.rb
```diff
@@ -20,6 +20,11 @@ def self.github_commit_uri(user, repo, sha = "$commit")
     "https://github.com/#{user}/#{repo}/commit/#{sha}"
   end
 
+  def enabled_for?(checker)
```
jrafanie: I like this but this should probably call `repo.enabled_for?(checker)` as checkers are enabled/disabled at the repo level.  @Fryguy thoughts?

jrafanie:
I like this, just a small question above... thanks @carbonin @imtayadeway  :heart:

jrafanie:
@Fryguy we should run flog before/after the PR and only then :scream: or :+1: 

It's nice to see that I modified complicated code but we really care about if I made it better or worse.

app/models/commit_monitor_branch.rb
imtayadeway: :+1: i had the same the same thought but we stopped ourselves here. i'd like to write some tests for this worker (though that could be a bigger investment of time), so didn't want to refactor too much until that was done.

@carbonin do you have bandwidth to look at this today?

carbonin: Yeah, definitely.

imtayadeway:
@jrafanie just added another commit with the suggested refactoring.

jrafanie:
Awesome, looks good.

jrafanie:
@imtayadeway @carbonin I added a screenshot of a sample PR demonstrating what it now looks like.  Try to add screenshots of the things that describe the change, also if you can't easily test via unit tests but through actual PR/Issues/etc.   :+1:

PR #83, opened 2015-06-26
Add a repos folder, but ignore its contents.

imtayadeway:
* Adds a `repos` folder so it does not need to be created in development steps
* Ignores anything in the `repos` folder so things added there for development are not tracked.
* Removes the corresponding step from the README

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@
 
 # Ignore production artifacts
 /public/assets
+
+# Ignore local repos folder
+/repos
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ problems.  Some of the things it does are:
 
 4. Add the sandbox app to the `repos` directory so it can be monitored:
    ```
-   mkdir repos
    git clone -o upstream git@github.com:miq-test/sandbox.git repos/sandbox
    ```
 


Fryguy:
Hmmm... This was one of the things I was not sure about in the original docs...I really don't think the repos should physically live in the same repo...would prefer if they lived externally (which is how we have it set up in deployment)

@brandondunne Thoughts?

Fryguy:
Actually, I've completely changed my mind and I love this repos directory idea as a default.  One of the things we want the bot to do is to autoclone the repos, which should make adding new ones easier.  Having the repos directory keeps it isolated from other repos that we might have locally.

However, does it make sense to have it configurable?  I'm thinking, keep the repos directory in the Rails root as the default, but perhaps allow overriding in a config file.  This might be useful in deployment if we want all the git repos to live elsewhere (say on an NFS share).  Having it configurable also allows different directories in prod/dev/test.

jrafanie:
> This might be useful in deployment if we want all the git repos to live elsewhere (say on an NFS share). Having it configurable also allows different directories in prod/dev/test.

Let's make it configurable when we know we need to do it.

I like this change.  :+1:

imtayadeway:
that sounds fun! anything to do be done at this time?

PR #88, opened 2015-07-02
Add pull request link to bugzilla ticket and change ticket status

carbonin:
bugzillachecker
The bugzilla checker now writes a comment with the PR link on a ticket when the ticket link is mentioned in a commit comment in a PR.
It will also change the bug status to ON_DEV if it was previously set to either NEW or ASSIGNED.

If the bug status was anything other then NEW, ASSIGNED, or ON_DEV when the bugzilla URI was seen in a commit message, the bot will add a needinfo flag and comment on the ticket as seen in the screenshot.


app/models/commit_monitor_branch.rb
```diff
@@ -39,6 +39,14 @@ def pr_number
     MiqToolsServices::MiniGit.pr_number(name) if pull_request?
   end
 
+  def github_pr_uri
+    unless pull_request?
+      raise ArgumentError, "Cannot retrieve pull request URI for non-pull request " \
```
imtayadeway: WDYT about raising a different error here or returning nil instead? I'm not sure about using `ArgumentError` here since it's being raised here for reasons unrelated to this method's arguments.

Also, I see there is a class level `.github_commit_uri` method in this class. You could make this a class level method to make this more uniform, or convert the other to be an instance method.

spec/models/commit_monitor_branch_spec.rb
```diff
@@ -71,6 +72,20 @@
     end
   end
 
+  context ".github_pr_uri" do
```
imtayadeway: it's conventional to use `#method_name` for instance methods and `.method_name` for class methods. but if you decide to change the implementation from the former to the latter then this can stay as is =)

i think i would use `describe` instead of `context` here. though they do the same thing, `context` is usually used for things like conditions or different arguments that affect the outcome.

app/workers/commit_monitor_handlers/commit/bugzilla_checker.rb
```diff
@@ -0,0 +1,112 @@
+module CommitMonitorHandlers
```
imtayadeway: it's hard to see what changed here! it looks like in `#process_commit` the worker will either do one thing or the other. what do you think about having a separate worker for each instead? then maybe any code they both need could be extracted into something they could both share. not sure about the feasibility of this....any thoughts?

app/models/commit_monitor_branch.rb
carbonin: I'm all for changing the error or returning nil, ArgumentError doesn't make much sense here.  Any suggestions for a different exception type?

I do think that this method makes more sense as an instance method because of the dependencies on object data.  It would be a bit cumbersome to pull that data out of an instance, then pass it back in to the class method.

The github_commit_uri method is a bit strange in that I think it could make sense as an instance method, but the only place it is being used is in the creation of the branch itself here:
https://github.com/ManageIQ/miq_bot/blob/master/app/models/commit_monitor_repo.rb#L20
I think it would be a bit odd to instantiate a branch object just to use that method to create a branch object.

spec/models/commit_monitor_branch_spec.rb
carbonin: Sounds good.  Also I think I'll change the name of the first test to something that makes more sense as this method doesn't actually take user and repo as parameters.

app/workers/commit_monitor_handlers/commit/bugzilla_checker.rb
carbonin: I like that idea.  I originally renamed the class because I realised it was doing more now than just commenting, but splitting them out seems like a cleaner solution.  Especially if we want to add more Bugzilla functionality in the future.

app/models/commit_monitor_branch.rb
Fryguy: Maybe just return nil if it's not a pull request.  Not sure if an exception is necessary.

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
```diff
@@ -0,0 +1,33 @@
...
+      def product
+        Settings.commit_monitor.bugzilla_product
+      end
+
+      def with_bug(bug_id)
+        return unless block_given?
+        MiqToolsServices::Bugzilla.call do
+          output = ActiveBugzilla::Bug.find(:product => product, :id => bug_id)
+          if output.empty?
+            logger.error "Unable to write for bug id #{bug_id}: Not a '#{product}' bug."
```
Fryguy: This error message doesn't make sense here...maybe raise an exception instead and let the caller handle it?  Another option is to just yield output.first, and let the caller handle nil itself.

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
```diff
@@ -0,0 +1,33 @@
...
+          else
+            bug = output.first
+            yield(bug)
+          end
+        end
+      rescue => err
+        logger.error "Unable to write for bug id #{bug_id}: #{err}"
```
Fryguy: Same...maybe just let the caller handle it.

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      def branch_valid?
+        if @branch.pull_request?
+          return true if self.class.handled_branch_modes.include?(:pr)
+        else
+          return true if self.class.handled_branch_modes.include?(:regular)
+        end
+        false
```
Fryguy: This seems to simplify to:

```ruby
def handled_mode?(mode)
  self.class.handled_branch_modes.include?(mode)
end

def branch_valid?
  (@branch.pull_request? && handled_mode?(:pr)) || handled_mode?(:regular)
end
```

app/workers/commit_monitor_handlers/commit/bugzilla_pr_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+      end
+
+      private
+
+      def update_bugzilla_status(bug_id)
+        with_bug(bug_id) do |bug|
+          add_pr_comment(bug)
+          update_bug_status(bug)
+          bug.save
```
Fryguy: Could you get into a position where save is called but nothing changes?  What happens?

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
```diff
@@ -0,0 +1,33 @@
+module CommitMonitorHandlers
```
Fryguy: The file location of this method is weird.  I feel like it should go a lot higher...maybe into /lib/bugzilla_worker_common.rb (or `shared` or `mixin`)

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
carbonin: I think that logic is incorrect for the case when the branch is a PR and the only handled mode is regular.  This seems like something I should just write a spec for.  I'm not that comfortable with rspec yet though ... Maybe @imtayadeway can help me out with this when he has some time.

imtayadeway: @carbonin absolutely! ping me when you're free

app/workers/commit_monitor_handlers/commit/bugzilla_pr_checker.rb
carbonin: It is possible, but nothing will happen.
https://github.com/ManageIQ/active_bugzilla/blob/master/lib/active_bugzilla/bug.rb#L25

app/workers/commit_monitor_handlers/commit/bugzilla_pr_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+      def bug_has_pr_uri_comment?(bug)
+        bug.comments.any? do |c|
+          c.text.include?(@branch.github_pr_uri)
+        end
+      end
+
+      def update_bug_status(bug)
+        bug_id = bug.id
+        bug_stat = bug.status
+        if bug_stat == "NEW" || bug_stat == "ASSIGNED"
```
jrafanie:
```ruby
if %w{NEW ASSIGNED}.include?(bug_stat)
```

jrafanie: Does this look better as a case statement? I would almost think we don't even want the ON_DEV branch of the case statement.  That's the normal case.

```ruby
def update_bug_status(bug)
  case bug.status
  when "NEW", "ASSIGNED"
    logger.info "Changing status of bug #{bug.id} to ON_DEV."
    bug.status = "ON_DEV"
  when "ON_DEV"
    # Good developer, you're doing it right!
  else
    logger.warn "Not changing status of bug #{bug.id} from #{bug.status}."
    bug.add_comment("Detected commit referencing this ticket while ticket status is #{bug.status}.")
    bug.flags["needinfo"] = "?"
  end
end

Comment (carbonin) [inline app/workers/commit_monitor_handlers/commit/bugzilla_pr_checker.rb]: We would still need a noop for ON_DEV so we could either keep the case like you wrote or make it an if with the else case being:

```ruby
elsif bug.status != "ON_DEV".
```

app/workers/commit_monitor_handlers/commit/bugzilla_common/bugzilla_common.rb
carbonin: For both of these cases, would it make sense to just return without a value and write a better log message?  That way the caller doesn't have to handle anything in a block that it doesn't want to be in anyway because we couldn't find the bug.

app/workers/commit_monitor_handlers/commit/bugzilla_pr_checker.rb
```diff
@@ -0,0 +1,74 @@
...
+      def bug_has_pr_uri_comment?(bug)
+        bug.comments.any? do |c|
+          c.text.include?(@branch.github_pr_uri)
+        end
+      end
+
+      def update_bug_status(bug)
+        bug_id = bug.id
+        bug_stat = bug.status
+        return if bug_stat == "ON_DEV"
```
jrafanie: :+1:

jrafanie:
LGTM.

PR #92, opened 2015-07-27
Move bot to issue manager worker

imtayadeway:
Fixes https://github.com/ManageIQ/miq_bot/issues/12


Fryguy:
Oh nice idea!  I hadn't considered just moving the runner over first!

PR #94, opened 2015-07-28
Make RSpec-Rails upgrades.

imtayadeway:
Recommended changes as detailed in:
https://www.relishapp.com/rspec/rspec-rails/docs/upgrade

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
 --format progress
+--require spec_helper
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -1,0 +1,2 @@
+test_class:
+  enabled_repos: []
--- a/spec/models/batch_entry_spec.rb
+++ b/spec/models/batch_entry_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'rails_helper'
 
 describe BatchEntry do
   include_examples "state predicates", :succeeded?,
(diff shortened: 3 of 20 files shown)


imtayadeway:
@jrafanie @Fryguy just want to check before updating/rebasing this PR if these changes are wanted. I know that these changes are opinionated and that some people prefer to eschew the `rails_helper`/`spec_helper` thing.... any thoughts?

Fryguy:
Yeah I'm on the fence, but mostly not really a fan of the changes, personally.  I don't really like the removal of the comments from the spec_helper.rb either as it creates noise when you try to upgrade that file (it's mostly generated as is).

I do like the `ActiveRecord::Migration.check_pending!` though.

Need to think about this a bit.

imtayadeway:
@Fryguy just did some work to at least bring this up to date/make it mergeable.  I killed the removal of the comments commit. I also had to make a couple of changes in a new commit to keep RSpec's verifying doubles happy - basically to ensure methods that are being stubbed are actually implemented on the object in question. They weren't necessarily real problems - in one case it was just an anonymous class, the other was rails-config's `Settings` object, but I personally like verifying doubles enough to deal with some of the minor inconveniences. Might be better ways to keep it happy here, but I didn't want spend too much time on it here.

Worth mentioning as this came up in conversation on gitter last week is the fact that RSpec-Rails 3 will generate the `rails_helper`/`spec_helper` more or less like this when you run the generation scripts (as I have done here). If this is something we do as a matter of practice upon updating RSpec it may be worth trying to come to a consensus on what way to go with this.

Anyway, no rush with this, and as I think I've said before I'm not that attached to all the changes brought about here, but there's some food for thought at least.

imtayadeway:
Going to close this while it's not going anywhere

Issue #95, opened 2015-07-28
Trello Bot


blomquisg:
### Goals of Using Trelllo
1. No one should have to touch the main roadmap board
   
   Team boards will drive the roadmap board.  And the roadmap board will largely be read-only.
2. Trello team boards should serve as the single place where devs can find things to work on.
   
   Bugzilla and Github will still be systems of record for bugs and pull requests, respectively.  But, Trello should be able to show the upcoming features as well as the bugs that are expected to be worked on for each sprint.
3. Product Managers should be able to see high level progress from roadmap.
   
   Trello team boards should feed the roadmap.  Items from various team board should be grouped together on the roadmap board by using consistent "tags" or labels across team boards.
4. Product Managers should be able to see specific team progress on team boards.
   
   Each team board should show the current status of work items for a sprint.
5. Team boards will have the same card lists:
   - **New**: The list of requirements from Product Management (or other sources) that have not yet been defined enough to scope
   - **Backlog**: The list of cards that have enough information to be scoped and have a decent time estimate.  This list is in rough priority order (most important on the top).
   - **Next**: The list of cards that could approximately build out the next one or two sprints.  This list is in rough priority order.
   - **In Progress**: The list of cards currently being worked on this sprint.
   - **Complete**: The list of cards completed during this sprint.  Cards can enter this list anytime during the sprint.
   - **Sprint &lt;NUMBER&gt;**: The list of cards that were completed for a given sprint.  Sprint cards can be archived and resurrected as need be.  We would likely keep four or five historical sprints in each team board at any time.  
### Trello Bot

Trello bot will help ensure that the team boards are in the correct state and display the correct information for each sprint.  However, it's still the responsibility of each team to make sure that the cards in Trello are kept up to date.
#### Requirements

**R1. Sync bugs from Bugzilla into Trello team boards**

  Each team board will have the same Bug cards for syncing bugs:
- **New Bugs**: A static card in the **New** list that contains the bugs assigned to the team but not yet assigned to any specific developer on the team.
- **Assigned Bugs**: A static card in the **Backlog** list that contains the bugs assigned to a specific developer on the team.
- **In Progress Bugs**: A static card in the **In Progress** list that contains the bugs in ON_DEV status and expected to be completed in that sprint.
- **Complete Bugs**: Static cards in the **Complete** and **Sprint &lt;NUMBER&gt;** lists that contain the lists of bugs in POST (and beyond) status.
  
[...]

PR #103, opened 2015-08-05
[WIP] Beginnings of Trello Bot

blomquisg:
[Trello Bot (Issue #95)](https://github.com/ManageIQ/miq_bot/issues/95) is a bot that will help us manage our usage of Trello in a few different ways:

1. Link bugs to Trello

  By linking bugs into Trello, developers will be able to see exactly which bugs they're responsible for in the same context as the feature work that they're responsible for.  In other words, it puts all of the work in one place.

  Also, it will give non-developers visibility into the **true** velocity of the team by incorporating the bugs that were fixed during a sprint as well as the feature work.

  As a bonus, developers will be able to check off bugs in card checklists to indicate that the bugs should be moved from "ASSIGNED" to "ON_DEV", or "ON_DEV" to "POST" directly from Trello without having to update Bugzilla manually.

2. Link Github to Trello

  As trello cards are mentioned in Github issues and PRs, trello cards can be updated with backlinks to those issues and PRs.  This improves traceability between Trello and Github.

3. Summary rollup boards

  Trello bot will also be able to create Rollup Boards to show over all objective summaries.  These rollups will allow non-developers to see the high-level picture of the progress of features across the workloads represented in Trello, while allowing developers to stay focused on the detailed views of their own Trello boards.


blomquisg:
@Fryguy, @imtayadeway, thoughts?

app/workers/trello_monitor.rb
```diff
@@ -0,0 +1,38 @@
+require 'bot/trello_helper'
+
+class TrelloMonitor
+  include Sidekiq::Worker
+  include Sidetiq::Schedulable
+  include MiqToolsServices::SidekiqWorkerMixin
+  sidekiq_options :queue => :trello_bot, :retry => false
```
Fryguy: This will have to be miq_bot in the end.

config/settings.yml
```diff
@@ -14,3 +14,100 @@ issue_manager:
...
+  oauth:
+    consumer_key:
+    consumer_secret:
+    oauth_token:
+    oauth_token_secret:
+  basic_auth:
+    developer_key:
+    member_token:
+  organization_id:
+  teams:
```
Fryguy: These should be defined elsewhere during deployment (I'm sure you already knew that...just clarfiying)

lib/bot/trello_helper.rb
```diff
@@ -0,0 +1,211 @@
...
+        ActiveBugzilla::Base.service
+      rescue NameError
+        service = ActiveBugzilla::Service.new(
+          Settings.bugzilla_credentials.bugzilla_uri,
+          Settings.bugzilla_credentials.username,
+          Settings.bugzilla_credentials.password
+        )
+        ActiveBugzilla::Base.service = service
+      end
+      ActiveBugzilla
```
Fryguy: This is already provided.  So, no need to rewrite.

Fryguy:
I think the trello stuff should live in a service object in miq_tools_services much like the rest of the services.  At that point the general setup is pretty straightforward.

Let's pair on this a little.  This is really awesome stuff!

Fryguy:
Also, thinking in a parallel model, the monitor should probably just poll something, look for changes, whatever, and then kick asynchronously kick off other work items.

app/workers/trello_monitor.rb
blomquisg: Oh, right ... I changed that in case I wanted to run it and not accidentally run the other workers...

I'm not altogether clear on how to do that, but I thought this might be an approach?

config/settings.yml
blomquisg: The teams?  Or, the auth keys?

I sorta like having the teams in this file.  It makes it really easy to define the mappings of teams to boards and bug groups, etc... 

Maybe you mean something else?

lib/bot/trello_helper.rb
blomquisg: K, I didn't dig very deep into the other helpers to find this ... I can replace this with a reference to another helper.

config/settings.yml
Fryguy: The teams.  I don't mind them being in configuration, just not in settings.yml.  They should be in settings.local.yml (which is not tracked) which is how we deploy the bot.

blomquisg: Ah, right ...

So, I have them locally in `development.local.yml`.  I can change this to be more generic (e.g., "team1, team2", etc...).

Fryguy:
A couple more things...

Don't put anything in lib/bot...that directory is on the chopping block.  It can live in lib directly, or eventually moved to miq_tools_services if that is more appropriate.

blomquisg:
> ...Don't put anything in lib/bot...
> ...the trello stuff should live in a service object in miq_tools_services...
> ...thinking in a parallel model, the monitor should probably just poll...

Ah, cool ... I can move that into lib for now.  Then, we can talk about what should move out to a separate service.

> Don't forget to the link the GH issue

Added to the PR summary.

blomquisg:
> Don't put anything in lib/bot

moved to `lib`

> [BZ service stuff] is already provided

using `MiqToolsServices::Bugzilla` now

> The teams...should be in settings.local.yml

updated `settings.yml` to just have examples

Gemfile
```diff
@@ -61,3 +61,7 @@ group :issue_manager do
   # Lock down dependency
   gem 'faraday', '~> 0.9.1'
 end
+
+group :trello do
```
Fryguy: Not sure we need a group at all.

PR #105, opened 2015-08-06
Refactor iceberg class pull request monitor

imtayadeway:
* Pulls the worker logic out of private methods and into public methods on the repo model and stateless services so that it can be more easily tested. 
* Adds specs for all the above
* Adds factory girl to aid in model tests

### DIFF
--- a/app/models/commit_monitor_repo.rb
+++ b/app/models/commit_monitor_repo.rb
@@ -71,13 +71,13 @@ class CommitMonitorRepo < ActiveRecord::Base
 
   def current_pr_branch_names
     with_git_service do |git|
-      break pull_requests.collect { |pr| git.pr_branch(pr.number) }
+      pull_requests.collect { |pr| git.pr_branch(pr.number) }
     end
   end
 
   def pull_requests
     with_github_service do |github|
-      break github.pull_requests.all
+      github.pull_requests.all
     end
   end
 
--- a/app/models/commit_monitor_repo.rb
+++ b/app/models/commit_monitor_repo.rb
@@ -64,4 +64,14 @@ class CommitMonitorRepo < ActiveRecord::Base
     repos = Settings.public_send(checker).enabled_repos
     fq_name.in?(repos)
   end
+
+  def pr_branches
+    branches.select(&:pull_request?)
+  end
+
+  def pull_requests
+    with_github_service do |github|
+      break github.pull_requests.all
+    end
+  end
 end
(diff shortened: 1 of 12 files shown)


app/models/commit_monitor_repo.rb
```diff
@@ -64,4 +64,24 @@ def enabled_for?(checker)
...
     fq_name.in?(repos)
   end
+
+  def pr_branches
+    branches.select(&:pull_request?)
+  end
+
+  def current_pr_branch_names
+    with_git_service do |git|
+      break pull_requests.collect { |pr| git.pr_branch(pr.number) }
```
Fryguy: break is not needed.

app/models/commit_monitor_repo.rb
```diff
@@ -64,4 +64,24 @@ def enabled_for?(checker)
...
+    end
+  end
+
+  def pull_requests
+    with_github_service do |github|
+      break github.pull_requests.all
```
Fryguy: Same

spec/workers/pull_request_monitor/pr_branch_record_spec.rb
```diff
@@ -0,0 +1,106 @@
+require "spec_helper"
+
+RSpec.describe PullRequestMonitor::PrBranchRecord do
+  describe ".create" do
+    it "has git create a local pr branch" do
+      repo = spy("CommitMonitorRepo")
```
Fryguy: Why spy vs double?  I never used spy before, and it seems inconsistent to introduce a new thing unless it has some different purpose.

Fryguy:
Looks nice!  One rubocop warning there, btw.

spec/workers/pull_request_monitor/pr_branch_record_spec.rb
imtayadeway: it has a distinct purpose which i find quite useful in tests like these where i'm trying to keep the setup phase under control. it's basically syntactic sugar for `double("CommitMonitorRepo").as_null_object`. i tend to use this quite a lot when creating doubles for collaborators whose use are just noisy implementation details which are not relevant to the particular test at hand. any thoughts?

imtayadeway:
@Fryguy thanks! i chose not to correct that warning as i found it helpful to give that object a 'name' by assigning it to an unused variable to better reveal its purpose.

app/models/commit_monitor_repo.rb
imtayadeway: nice! did not realize that would work

spec/workers/pull_request_monitor/pr_branch_record_spec.rb
Fryguy: My only thought was that spies are really for after-the-fact expectation checks, so it seems weird to use them as_null_object, but it is clever :smile:  I'm fine with it for now.

PR #106, opened 2015-08-14
Rename models

imtayadeway:
don't think there's an issue for this, but I'm pretty sure we talked about this change:

* `CommitMonitorRepo` => `Repo`
* `CommitMonitorBranch` => `Branch`

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ underscores replaced with hyphens.
 
 7. Add the test repository to the database:
    ```
-   bundle exec rails runner 'CommitMonitorRepo.create_from_github!("miq-test", "sandbox", Rails.root.join("repos", "sandbox"))'
+   bundle exec rails runner 'Repo.create_from_github!("miq-test", "sandbox", Rails.root.join("repos", "sandbox"))'
    ```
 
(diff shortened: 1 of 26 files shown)


Fryguy:
IT'S LIKE YOU'RE READING MY MIND... :open_mouth: 

The only thing I was going to do different was name them GitRepo and GitBranch.  Thoughts?

Fryguy:
And no, I don't think we had talked about it...But I was literally going to change this right after my batch support changes :smile:

imtayadeway:
@Fryguy nice! hehe

I think what I like about `Repo`/`Branch` is that it's natural to say that a `repo` has many `branches`, and everything Just Works :tm: etc..

What I like about `GitRepo`/`GitBranch` is that it's more immediately obvious what these things are modeling from the outside without any further context. So I'm kinda on the fence :grinning:

Fryguy:
Been thinking about it and I'm good with Repo / Branch.  But, alas, it has conflicts

imtayadeway:
@Fryguy just rebased and incorporated some new stuff along the way.

PR #107, opened 2015-08-18
Add batch support

### LINKED ISSUE #58: Batch support for bulk commenting
I have a design for this, which I've shared with @jrafanie.  I'll have to elaborate here or I'll just make a PR.

Fryguy:
Resolves #58 .

This PR includes a number of things to fully support batching of jobs.
- Add BatchJob model which has_many BatchEntry.
- Add a "batch" type to the CommitMonitor.  Ultimately, I'm considering throwing away all the "types", but for now this fits in with the existing format.
- Added GithubPrCommenter as an initial batch worker.  I'm not thrilled with the name, but we can change it later.
- Move the GemfileChecker as a batch entry under GithubPrCommenter
- Pulled out concern mixins for BatchJobWorkerMixin, and BatchEntryWorkerMixin.  I didn't yet apply these to the other class, but will after this PR is merged.  Again, not thrilled with the names, but it can be changed
- Lots of specs and lots of new spec/support methods for testing workers.


Fryguy:
cc @imtayadeway @brandondunne @jrafanie If you care to review.  It's pretty long though...I tried breaking up commits for easier reading, but it's still a lot.

Fryguy:
TODO: I might collapse the 3 migrations.  No need to have 3 separate migrations in the same PR even though I slowly worked my way up to them when developing.

app/models/batch_job.rb
```diff
@@ -0,0 +1,63 @@
...
+  def on_complete_class=(klass)
+    super(klass.to_s)
+  end
+
+  def finalizing?
+    state == "finalizing"
+  end
+
+  def expired?
+    expires_at && Time.now > expires_at
```
imtayadeway: could this be `Time.zone.now`?

Fryguy: I guess, but I don't understand why that matters (also don't understand why Rubocop cares).  Can you explain?

Fryguy:
Also, PullRequestMonitor and CommitMonitor are all kinds of messed up right now after merging #105 and #106 ...need to fix on master.  I have some fixes locally, but not 100% there yet.

app/models/batch_job.rb
imtayadeway: Just wondering if those tests for `#expired?` could fail under different conditions....is `Time.zone` ever likely to change?

imtayadeway:
@Fryguy oops! anything I can do?

As far as this PR goes, it is indeed big but LGTM :+1:

Fryguy:
This is ready to go

PR #115, opened 2015-09-01
Print the ruby version in rubocop comments

### LINKED ISSUE #114: Bot should report what ruby version it's using in it's rubocop comment
Sometimes the bot ruby version doesn't match the developer and the warnings about invalid syntax make very little sense unless you know it's using an older/newer ruby.

screen shot 2015-09-01 at 10 21 13 am

screen shot 2015-09-01 at 10 22 15 am

jrafanie:
Sometimes, rubocop will be running on a different ruby version than
the developer who opened the pull request.

This leads to very :confused: developers as rubocop could report
syntax errors such as %i (added in 2.0) if the bot is using ruby 1.9.3.

Printing the ruby version may help clarify why it's invalid syntax.

Fixes #114

### DIFF
--- a/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
@@ -43,7 +43,7 @@ class CommitMonitorHandlers::CommitRange::RubocopChecker::MessageBuilder
   end
 
   def header
-    header1 = "Checked #{"commit".pluralize(commits.length)} #{commit_range_text} with rubocop #{rubocop_version} and haml-lint #{hamllint_version}"
+    header1 = "Checked #{"commit".pluralize(commits.length)} #{commit_range_text} with ruby #{RUBY_VERSION}, rubocop #{rubocop_version}, and haml-lint #{hamllint_version}"
 
     file_count    = results.fetch_path("summary", "target_file_count").to_i
     offense_count = results.fetch_path("summary", "offense_count").to_i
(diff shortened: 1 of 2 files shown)


jrafanie:
@Fryguy @brandondunne @blomquisg hope this helps a little

blomquisg:
:+1: Thanks @jrafanie

brandondunne:
Totally in favor of the change, but should we use `RUBY_DESCRIPTION` instead?  I know it's more wordy, but it also includes the patch level and platform.

jrafanie:
I guess platform and patch level is useful, I don't know if it would information overload though:

```
jruby 1.7.18 (1.9.3p551) 2014-12-22 625381c on Java HotSpot(TM) 64-Bit Server VM 1.8.0_31-b13 +jit [darwin-x86_64]
ruby 1.9.2p330 (2014-08-07) [x86_64-darwin13.4.0]
ruby 1.9.3p547 (2014-05-14) [x86_64-darwin13.4.0]
ruby 2.1.7p400 (2015-08-18 revision 51632) [x86_64-darwin13.0]
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin13]
```

Fryguy:
Tests fail because there are things that check the text in the rubocop commenter...they can probably be changed to avoid unnecessary things like this.

Also, if you use RUBY_DESCRIPTION, it starts with the word ruby, but you also have ruby in the string, so it would appear twice.

matthewd:
I don't think we're likely to encounter any situation where more than the minor version (2.0 vs 2.1) makes a difference anyone cares about.

jrafanie:
Ok, rebased and fixed the tests, kept it as RUBY_VERSION.  I don't think the patchlevel platform is useful enough to warrant the extra noise.

jrafanie:
@blomquisg @Fryguy this is ready to go

PR #123, opened 2015-09-03
Add checker for bad migration timestamps

### LINKED ISSUE #113: Check for invalid dates in migration names
An invalid date in a migration file name is considered a valid value, but then migrations take the form of numerics instead of dates from then on causing weirdness.  This check would prevent that.

Fryguy:
Resolves #113 

@jrafanie Please review.

### DIFF
--- a/app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
+++ b/app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
@@ -1,0 +1,44 @@
+require 'time'
+
+module CommitMonitorHandlers::Batch
+  class GithubPrCommenter::MigrationDateChecker
+    include Sidekiq::Worker
+    sidekiq_options :queue => :miq_bot
+
+    include BatchEntryWorkerMixin
+    include BranchWorkerMixin
+
+    def perform(batch_entry_id, branch_id, _new_commits)
+      return unless find_batch_entry(batch_entry_id)
+      return unless find_branch(branch_id, :pr)
+
+      complete_batch_entry(bad_migrations.any? ? {:result => new_comment} : {})
+    end
+
+    private
+
+    def new_comment
+      ":red_circle: Bad migration #{"date".pluralize(bad_migrations.length)}: #{bad_migrations.join(", ")}"
+    end
+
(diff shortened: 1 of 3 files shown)


app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
```diff
@@ -0,0 +1,44 @@
+require 'time'
+
+module CommitMonitorHandlers::Batch
+  class GithubPrCommenter::MigrationDateChecker
```
jrafanie: Is there a reason to use this condensed style?  It seems to be more hassle with autoload and constants on each namespace than useful.

app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def valid_timestamp?(ts)
+      Time.parse(ts)
+    rescue ArgumentError
+      false
+    end
+
+    def diff_files_for_commit_range
+      branch.repo.with_git_service do |git|
+        git.diff_details(*commit_range).keys
```
jrafanie: I like this, does it make sense to have a helper method for all commit/commit range handlers that gets the diff_details based on the commit/commit range?  It seems like all of them might want to look at the commit data.

jrafanie:
LGTM, other than the namespacing question.

app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
Fryguy: I didn't want 5 levels of indent...it's all going to change anyway when I drop the "type" part

app/workers/commit_monitor_handlers/batch/github_pr_commenter/migration_date_checker.rb
Fryguy: Yes I plan to in the future...wanted a few in place so a pattern could emerge. https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

PR #125, opened 2015-09-16
Remove legacy bot from lib

imtayadeway:
### Move IssueManager into main app

A continuation of work started in #92 to bring the legacy bot code
into the main app. This commit:

* Updates the Gemfile to move the `:issue_manager` group gems into the
  top level
* Updates the `Procfile.tmpl` to no longer have an entry for the now
  non-existent legacy bot log
* Moves the `GithubApi` library into `app/models`
* Moves `IssueManager` into `app/services`
* Updates `IssueManagerWorker` to no longer explicitly require
  `IssueManager` since it will now be autoloaded
* Updates any legacy bot code to use the Rails logger
* Removes the `lib/bot` directory from the repository
* Adds `config/issue_manager.yml` to the gitignore
* Adds a namespace to `GithubApi::Repo` where needed to avoid clashing
  with the top level `Repo` in `app/models`.

### Rehash legacy specs for IssueManager

This commit reworks some of the forgotten/failing tests that resided in
`lib/bot/spec`. An attempt to tweak them in order to get them to green
was made, but owing to the extensive stubbing and setup in general it
became too hard. I decided to rewrite them using the same
specs (i.e. they test the same thing, but differently).

In particularly, it:

* Introduces an `IssueManager.build` method which replaces the `.new`
  method. This allows for the constructor to be much simpler and for a
  lot of the details to be abstracted away, which simplifies the tests
  somewhat.
* Updates everything to use `IssueManager.build` where appropriate.
* Removes VCR. It would be nice to bring this back at
  some point.
* Avoids stubbing Octokit (object we don't own), and leverages some
  verifying doubles to give more confidence

Some unfortunate things that further refactoring will hopefully eliminate:

* Stubbing `File.write` to avoid changing the world - `IssueManager`
  persists data to a local yaml file.
* Stubbing `YAML.load_file` for the same reason.
* Still too much stubbing and especially nested stubs in the setup for
  `IssueManager` specs
* Reduces test coverage (only in theory - in practice these tests
  weren't being run/maintained) around the network boundary. It would be
  nice to add some integration tests here

### Some thoughts on this PR and where to go next
* not crazy about creating `app/services`, open to suggestions for where to put `issue_manager.rb`
* similarly with the `GithubApi` stuff in `app/models` - though it does seem somewhat appropriate as most of the things in that library are models
* is the `IssueManager` really an _issue_ manager? I found this confusing at first. Perhaps rename to `NotificationManager` or something similar


Fryguy:
> is the IssueManager really an issue manager? I found this confusing at first. Perhaps rename to NotificationManager or something similar

I had thought of renaming this to NotificationMonitor, so I agree

> not crazy about creating app/services, open to suggestions for where to put issue_manager.rb

I think it should go in app/models or lib

> similarly with the GithubApi stuff in app/models

The GitHubApi stuff should go away along with Octokit.  That was how we originally wrote it until we found the github_api gem that we use through miq_tools_services.  I am ok with keeping it in a first pass PR like this, though.  I think it belongs in lib, but honestly, that's not a big deal.

> IssueManager persists data to a local yaml file.

Ugh yeah, we have to move that into the database at some point.

Fryguy:
Did you actually test this against the a repo?

imtayadeway:
> The GitHubApi stuff should go away along with Octokit. That was how we originally wrote it until we found the github_api gem that we use through miq_tools_services. I am ok with keeping it in a first pass PR like this, though. I think it belongs in lib, but honestly, that's not a big deal.

Yeah, I originally attempted that in this PR, but it started to get messy, so I'd like to tackle that in a follow up PR. After integrating the logging and getting the tests working I felt that the diff was telling me to stop there :)

> I had thought of renaming this to NotificationMonitor, so I agree

Cool, I'll either do that here or in a follow up PR

> Did you actually test this against a real repo?

I did some testing on sandbox, but that was a while ago so let me do some more testing before this gets merged to answer that more confidently

Fryguy:
Actually, I was considering GithubNotificationMonitor

imtayadeway:
@Fryguy just confirmed - all tests pass, app boots up and will add/remove labels from the sandbox. That's after I added another commit just to move the `IssueManager` into `app/models`.

I'll handle the renaming of the `IssueManager`, elimination of the `GitHubApi`/Octokit stuff and dealing with `IssueManager`'s persistence layer in some follow up PRs, if that's OK. Anything else to do here?

PR #129, opened 2015-09-23
Let ActiveRecord serialize the column to avoid:

bdunne:
2015-09-23T19:08:46Z 875 TID-mfjfo WARN: Attribute was supposed to be a Array, but was a String. -- "---\n- b912c2c82de4463ddf8a926d1ad3c04b5b726d5a\n"


bdunne:
@Fryguy @matthewd Please review

PR #130, opened 2015-10-07
Fix issue where TravisBuildKiller grouped all non-PRs together.

### LINKED ISSUE #52: TravisBuildKiller treats all branches and tags the same, but shouldn't
If I update a branch like botvinnik, then the current master branch build gets killed.  This is because the code is lumping all non-PRs together, and probably should group them by branch/tag name instead.

Fryguy:
Fixes #52 

@jrafanie @bdunne Please review
@matthewd This should fix the issue with capablanca.


jrafanie:
:heart:

PR #132, opened 2015-10-08
Kill Travis builds that are running for over 1 hour

bdunne:
Sometimes Travis builds get stuck and Travis doesn't kill them.  This can result in builds running for 18+ hours before someone notices and manually kills them.  Since that worker was tied up, the Travis backlog continues to grow.

### DIFF
--- a/app/workers/travis_build_killer.rb
+++ b/app/workers/travis_build_killer.rb
@@ -30,13 +30,10 @@ class TravisBuildKiller
 
   def kill_builds
     repo.with_travis_service do |travis|
-      builds_to_cancel = travis.builds
-        .take_while { |b| b.pending? || b.canceled? }
-        .reject(&:canceled?)
-        .group_by { |b| b.pull_request_number || b.branch_info }
-        .flat_map { |_key, builds| builds[1..-1] }
+      pending_builds = travis.builds.take_while { |b| b.pending? || b.canceled? }.reject(&:canceled?)
 
-      builds_to_cancel.each do |b|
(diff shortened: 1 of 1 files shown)


bdunne:
@jrafanie @Fryguy Please review.

cc @kbrock

kbrock:
@bdunne cool.

Should we put something different around the vmdb build? that seems to be teetering around 1 hour mark. Maybe bump that up to 1 1/2 or 2?

kbrock:
misspoke. vmdb takes 30 minutes.
Do we want to not kill master? or can we restart those?

bdunne:
re-working... this should actually look at the job, not the build

app/workers/travis_build_killer.rb
```diff
@@ -47,6 +45,12 @@ def kill_builds
           b.cancel
         end
       end
+
+      long_running_jobs.each do |b|
+        next unless b.running?
```
Fryguy: How is this line ever not running?

bdunne: If we cancel it above in the `out_dated_builds`

Fryguy: But the first thing you do on https://github.com/ManageIQ/miq_bot/pull/132/files#diff-d80c65f3a5fe96f44114c253eb74c918R35 is to only select the running ones.

Fryguy: Oh, if you cancel it, in between?...that's really confusing...perhaps instead, do the `long_running_jobs =` line after you do the outdated builds?

Fryguy: Or remove the outdated_builds from the long_running_jobs ?

bdunne:
@Fryguy updated.

PR #133, opened 2015-10-14
Rename IssueManager -> GithubNotificationMonitor

imtayadeway:
Likewise, the IssueManagerWorker has become
GithubNotificationMonitorWorker, and all references updated.

The name no longer reflected the responsibility of this class -
notifications and issues are two completely separate things in the
Github domain, and the class deals with Github notifications.

Note: This has a couple of destructive changes that may affect production:

1. `Settings.issue_manager` is now `Settings.github_notification_monitor`, so the the settings.yml file will need to be updated accordingly
2. Renames the yaml file that persists the notification monitor data

I will happily back out either if this causes any unnecessary pains.

This is a follow up to https://github.com/ManageIQ/miq_bot/pull/125 - more to follow.


Fryguy:
:+1: I'll merge after we deploy the current master, so we can defer the destructive changes.

imtayadeway:
@Fryguy just realized I missed a reference in the README and the `settings.yml` template so added another commit to cover those.

Fryguy:
@imtayadeway Can you rebase...we had to make some other fixes, but I want to get this in. Thanks!

imtayadeway:
@Fryguy done! =D

PR #138, opened 2015-12-03
Create a slow queue for tasks that require locking on git or Github

bdunne:
img_20151203_134528552_hdr

### DIFF
--- a/Procfile.tmpl
+++ b/Procfile.tmpl
@@ -1,4 +1,5 @@
 thin:            bundle exec rails s -p $PORT
 sidekiq:         bundle exec sidekiq -q miq_bot
+sidekiq_glacial: bundle exec sidekiq -q miq_bot_glacial
 travis_listener: bundle exec rails runner lib/travis_event/listener.rb
 #rails:           tail -f -n 0 log/development.log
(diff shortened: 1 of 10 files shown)


bdunne:
@Fryguy Please review

bdunne:
cc @matthewd @jrafanie

Procfile.tmpl
```diff
@@ -1,4 +1,5 @@
 thin:            bundle exec rails s -p $PORT
 sidekiq:         bundle exec sidekiq -q miq_bot
+sidekiq:         bundle exec sidekiq -q miq_bot_glacial
```
matthewd: Doesn't this name (lhs) need to be unique?

Fryguy: It doesn't *have* to be, but it makes reading the log much harder if it's not.  @bdunne Please change.

jrafanie: Yes, good call

Fryguy:
:+1: aside from the foreman name thing.

jrafanie:
I can't wait to see glacial log messages

jrafanie:
Looks good to me.

PR #144, opened 2015-12-09
Retain only new commits for non-PR branches.

jrafanie:
The statistics hash is retained through the life of one run of the
commit monitor, which can traverse 100+ branches, so all commits and
new commits for all branches is a lot of data.

We currently only care about "master" so we can compare PRs against
any new commits to "master".  Using Branch#pull_request? will allow
us to use non-master branches at some point.

Additionally, we don't ever use statistics[:all_commits] so don't
store it.


jrafanie:
cc @Fryguy @bdunne

bdunne:
:+1: Nice

PR #145, opened 2015-12-10
Rubocop shouldn't block on git

bdunne:
- Guarantee that we're getting the correct ruby / rubocop when linting
- Don't require rubocop in every process (this is the only place it's called)
- Upgrade to the latest rubocop


bdunne:
@Fryguy @matthewd Please review

lib/linter.rb
```diff
@@ -0,0 +1,4 @@
+module Linter
+  linters_path = Rails.root.join("lib", "linters")
+  Dir.glob(linters_path.join("*.rb")).each { |file| require_relative(file) }
```
matthewd: If it's in `autoload_paths`, do we need to do this?

Gemfile
```diff
@@ -41,7 +41,7 @@ gem 'travis', '~>1.7.6'
 gem 'awesome_spawn'
 gem 'default_value_for'
 gem 'more_core_extensions', "~> 2.0.0", :require => 'more_core_extensions/all'
-gem 'rubocop', '>= 0.27.1'
+gem 'rubocop', '>= 0.27.1', :require => false
 gem 'haml-lint', '~> 0.13.0'
```
matthewd: I think this ultimately requires rubocop too (https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/linter/rubocop.rb#L2), so might need the same treatment before we gain anything.

lib/linter.rb
bdunne: @matthewd It still wasn't being found until this was added.

bdunne:
cc @jrafanie

lib/linter.rb
matthewd: Oh, because the directory is `linters` but the constant is `Linter`?

bdunne:
@Fryguy Please re-review, I kept the shelling out part as discussed.

lib/linter/rubocop.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def run
+      require 'tempfile'
+      result = Dir.mktmpdir do |dir|
+        @work_dir = File.join(dir, "rubocop")
+        Dir.mkdir(@work_dir)
+        collect_files
+
+        options = {:format => 'json'}
+        @branch.logger.info("Executing: #{AwesomeSpawn.build_command_line('rubocop', options)}")
```
Fryguy: I just realized you don't have the .rubocop.yml from the repo, so what's it going to use?

bdunne: @Fryguy I added it on line 36

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -65,9 +48,6 @@ def extract_haml_files(diff_details)
   def linter_results(cmd, options = {})
     require 'awesome_spawn'
 
-    # rubocop exits 1 both when there are errors and when there are style issues.
-    #   Instead of relying on just exit_status, we check if there is anything
-    #   on stderr.
```
Fryguy: Why is this comment removed?

Fryguy: OH I see...it's moved down below...nvm

app/models/branch.rb
```diff
@@ -75,4 +75,22 @@ def write_github_comment(header, continuation_header = nil, message = nil)
...
+    merge_head  = "origin/master" # TODO: should be a column
+    rugged_repo = Rugged::Repository.new(repo.path.to_s)
+    rugged_repo.diff(merge_head, branch_ref_name)
+  end
+
+  def git_branch_ref
+    require 'rugged'
+    rugged_repo = Rugged::Repository.new(repo.path.to_s)
+    rugged_repo.references["refs/#{branch_ref_name}"]
+  end
```
Fryguy: Note that we should move these rubocop methods out into a service on miq_tools_services...but that can be a follow up.

lib/linter/rubocop.rb
```diff
@@ -0,0 +1,56 @@
...
+      diff             = @branch.git_diff
+      files_changed    = diff.deltas.collect { |delta| delta.try(:new_file).try(:[], :path) }.compact
+      files_to_rubocop = filtered_files(files_changed)
+
+      (files_to_rubocop + [".rubocop.yml"]).each do |file|
+        blob = diff.owner.blob_at(ref.target.oid, file)
+        next unless blob
+        temp_file = File.join(@work_dir, file)
+        FileUtils.mkdir_p(File.dirname(temp_file))
+        File.write(temp_file, blob.content)
```
Fryguy: This kind of stuff would live in a service class too in a method like `extract_blob_to_file`

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -19,15 +19,10 @@ def perform(branch_id, _new_commits)
...
-
     unmerged_results = []
+    unmerged_results << Linter::Rubocop.new(branch).result
 
-    files = extract_ruby_files(diff_details)
-    if files.any?
-      unmerged_results << linter_results('rubocop', :format => 'json', nil => files)
-    end
-
+    diff_details = diff_details_for_merge
```
Fryguy: I guess you don't really get any gains from this PR until the haml linter no longer needs to go to a git checkout either?

bdunne: Do you want me to do that in this PR?  I was going to keep them separate.

Fryguy: Yeah separate is fine, it just occurred to me when reading this.

bdunne:
@Fryguy Anything you wanted me to change?

bdunne:
@Fryguy rebased

PR #150, opened 2015-12-18
Prevent pull request monitor from blocking on git

bdunne:
Based on changes in https://github.com/ManageIQ/miq_bot/pull/145


bdunne:
cc @Fryguy @jrafanie

PR #151, opened 2016-01-04
No longer need to pass git service when pruning pr branches

bdunne:


bdunne:
@Fryguy Please review

PR #157, opened 2016-02-17
[WIP] Feature: Auto-add WIP label by title

chrisarcand:
This is a quick stab at auto-assigning the WIP label to any issues titled as WIP. It's pretty simplistic and doesn't include any specs - honestly, if we want to wait for them I'd rather do some refactoring first as trying to add them was a nightmare without changing a ton of unrelated specs and I give up for now for such a small change.

Note that one limitation here is that issue/PR renames are _not_ included in GitHub's Notification API, meaning the bot doesn't actively add/remove the label unless some other notification has triggered it. That is, it'll be labeled upon first opening the PR, but won't be labeled if someone changes the title mid way through until someone comments, it's closed, merged, etc.

### DIFF
--- a/app/models/github_notification_monitor.rb
+++ b/app/models/github_notification_monitor.rb
@@ -53,9 +53,18 @@ class GithubNotificationMonitor
   end
 
   def process_issue_thread(issue)
+    process_issue_title(issue)
     process_issue_comment(issue, issue.author, issue.created_at, issue.body)
     issue.comments.each do |comment|
       process_issue_comment(comment.issue, comment.author, comment.updated_at, comment.body)
+    end
+  end
+
+  def process_issue_title(issue)
+    if issue.title_indicates_wip?
+      issue.add_labels([GitHubApi::Label.new(nil, "wip", nil)]) unless issue.applied_label?("wip")
+    else
+      issue.remove_label("wip") if issue.applied_label?("wip")
     end
   end
 
--- a/app/models/git_hub_api/issue.rb
+++ b/app/models/git_hub_api/issue.rb
@@ -62,6 +62,10 @@ module GitHubApi
       end
     end
 
+    def title_indicates_wip?
+      !!(@title =~ /\[WIP\]/i)
+    end
+
     private
 
     def update(options)


chrisarcand:
Ironically, labeling this as WIP...

Fryguy:
Nice.

> Note that one limitation here is that issue/PR renames are not included in GitHub's Notification API

That stinks.  When I first just read the code, I got excited thinking it did :/  I think this could be handled in the Pull Request monitor, which is hitting the GitHub API anyway.

One other thing this PR doesn't address is when someone calls `@miq-bot add_label wip` or `@miq-bot rm_label wip`, I think it should update the PR title.  Otherwise they will get in this weird position where they haven't changed the title, the bot will add the wip label, and then when the title is examined to not have the WIP text, it will remove the wip label. (If that makes sense)

app/models/github_notification_monitor.rb
```diff
@@ -53,12 +53,21 @@ def process_notification(notification)
...
+    process_issue_title(issue)
     process_issue_comment(issue, issue.author, issue.created_at, issue.body)
     issue.comments.each do |comment|
       process_issue_comment(comment.issue, comment.author, comment.updated_at, comment.body)
     end
   end
 
+  def process_issue_title(issue)
+    if issue.title_indicates_wip?
+      issue.add_labels([GitHubApi::Label.new(nil, "wip", nil)]) unless issue.applied_label?("wip")
```
Fryguy: I wonder if this would be cleaner to call the `add_labels` method below with something like `add_labels("wip", issue.author, issue)`

Fryguy:
@chrisarcand I think this is a great first pass if we could also add in the auto-title changing thing I mentioned [here](https://github.com/ManageIQ/miq_bot/pull/157#issuecomment-185387347).

I thin moving it to the PR Monitor could come later.

chrisarcand:
@Fryguy Fair requests. Leaving this PR open for now but I'll likely redo it for the most part. A lot of the things we both want could be done so much easier with a bit of cleanup before this is added, IMO :(

Fryguy:
@chrisarcand What in particular?  Just wondering, I am not opposed to fixing stuff :smile: 

Note that the entirety of the GitHubApi class should eventually :sparkles: go away :sparkles: and instead we should use the gem, but that might be a bit of work.

Fryguy:
Closing in favor of #220, since that one uses the PR monitor instead and will be more reliable.

PR #160, opened 2016-03-01
Remove comment and needinfo flag when BZ is referenced after ONDEV state

carbonin:
This will avoid issues like the one that happenend here https://bugzilla.redhat.com/show_bug.cgi?id=1296248 where the bot comments and sets needinfo when there was a bad rebase.


Fryguy:
:+1:

PR #161, opened 2016-03-03
Fix rugged diffs

bdunne:
Previously we were diffing between the tips of the branches and getting too many deltas.  We should be diffing between the merge_base and the tip of the PR.

Helpful link... https://github.com/libgit2/libgit2/issues/1679

### DIFF
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -86,8 +86,7 @@ class Branch < ActiveRecord::Base
   end
 
   def git_diff
-    merge_head  = "origin/master" # TODO: should be a column
-    rugged_repo.diff(merge_head, branch_ref_name)
+    rugged_repo.diff(git_merge_base, branch_ref_name)
   end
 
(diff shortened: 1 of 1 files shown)


bdunne:
@carbonin @Fryguy Please review

app/models/branch.rb
```diff
@@ -86,21 +86,23 @@ def branch_ref_name
   end
 
   def git_diff
-    require 'rugged'
     merge_head  = "origin/master" # TODO: should be a column
-    rugged_repo = Rugged::Repository.new(repo.path.to_s)
-    rugged_repo.diff(merge_head, branch_ref_name)
+    merge_base  = rugged_repo.merge_base(merge_head, branch_ref_name)
```
matthewd: This looks suspiciously similar to `git_merge_base` below

bdunne: good point... cleaning up

PR #163, opened 2016-03-07
Leverage merge_target column

bdunne:


bdunne:
@Fryguy Please review.  That should be the last one.

PR #164, opened 2016-03-10
Fix rugged diffs for rubocopping

bdunne:
Add GitService::Branch wrapper to simplify branch methods
Add GitService::Diff wrapper to simplify walking changes
Update Linter::Rubocop to leverage the new GitService classes

### DIFF
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -80,27 +80,7 @@ class Branch < ActiveRecord::Base
     end
   end
 
-  def branch_ref_name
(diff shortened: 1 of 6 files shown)


bdunne:
@Fryguy Please review

bdunne:
cc @carbonin

bdunne:
The "Don't suppress exceptions" warning should be ignored.  This pattern is mostly copied from `blob_at` in Rugged::Repository and we actually want to suppress the exception here.

PR #171, opened 2016-03-31
Don't retry to rubocop an unmergable PR

bdunne:
Currently, when a rubocop linter tries to run against an unmergable PR, Rugged fails to create an index and raises `Rugged::IndexError`.  That error is raised and Sidekiq puts the job into retry.  We should not retry the job, since a rebase and push from the user should trigger a new rubocop run to be scheduled.


bdunne:
@Fryguy Please review

Issue #175, opened 2016-05-08
Scripts to Rull them All


juliancheal:
We should implement https://github.com/github/scripts-to-rule-them-all

We already kinda have something like this in https://github.com/ManageIQ/manageiq/tree/master/bin

miq_bot assign @juliancheal

juliancheal:
@miq-bot assign @juliancheal

juliancheal:
Kinda duplicate/follow on from https://github.com/ManageIQ/miq_bot/issues/156

chriskacerguis:
@miq-bot set_milestone Bot v2

PR #176, opened 2016-05-08
MiQ-Bot now with added update and setup

juliancheal:
For consistancy and ease of use basically a copy of MangeIQ's
setup and update commands.

Also the beginings of following along to https://github.com/github/scripts-to-rule-them-all

Fixes issue #175

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,0 +1,56 @@
+#!/usr/bin/env ruby
+require 'pathname'
+require 'optparse'
+
+
+options = {:do_tests => true, :do_db => true}
+OptionParser.new do |opts|
+  opts.banner = "Usage: setup [options]"
+  opts.on("-d", "--no-db", "Do not prepare db") do
+    options[:do_db] = false
+  end
+  opts.on("-t", "--no-tests", "Do not perform tests") do
+    options[:do_tests] = false
+  end
+  opts.on("-h", "--help", "Displays this help") do
+    puts opts
+    exit
+  end
+end.parse!
+
+
+# path to your application root.
+APP_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -0,0 +1,56 @@
...
+
+  puts "\n== Copying sample files =="
+  unless File.exist?("config/database.yml")
+    system "cp config/database.tmpl.yml config/database.yml"
+  end
+  unless File.exist?("Procfile")
+    system "cp Procfile.tmpl Procfile"
+  end
+  unless File.exist?("config/settings/development.local.yml")
+    system "cp config/settings.yml config/settings/development.local.yml"
```
bdunne: Why create this file if it contains no changes?

bin/update
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env ruby
+require 'pathname'
+
+# path to your application root.
+APP_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
+
+Dir.chdir APP_ROOT do
+  puts "== Installing dependencies =="
+  system "bundle update"
+  system "bower update --allow-root -F --config.analytics=false"
```
bdunne: I don't think we use bower in this app.

juliancheal: Cool, I'll remove that. :)

bin/setup
juliancheal: Sorry, I'm not sure what you mean?

bdunne: I don't think it makes sense to copy `config/settings.yml`.  I would only expect that file to contain intentional differences.

bin/update
himdel: ..but we probably *will* need npm :) (JS linters..)

PR #178, opened 2016-05-11
Haml linter

bdunne:
Move Haml linter to a module like rubocop


bdunne:
@Fryguy Please review

PR #190, opened 2016-06-01
Use rugged to determine branch mergeability

bdunne:
Command line merging is painfully slow and the volume is becoming a problem


bdunne:
@Fryguy Please review

PR #193, opened 2016-06-15
Remove the BugzillaPrChecker worker

carbonin:
This worker was causing more trouble than it was worth by commenting on BZs when someone pushed a bad rebase to a PR.

This situation would cause many unrelated commits to be included in a PR branch which would trigger the worker to comment on all BZ tickets referenced in those commits.

These comments would reference an unrelated PR and could not be removed from the BZ.

See the following BZs for examples:
https://bugzilla.redhat.com/show_bug.cgi?id=1340247
https://bugzilla.redhat.com/show_bug.cgi?id=1341291
https://bugzilla.redhat.com/show_bug.cgi?id=1344112


carbonin:
@Fryguy please review
/cc @imtayadeway I think we've now removed almost all of the initial work we did on the bot :cry:

bdunne:
:-1: I actually like this feature.  I find that I interact with Bugzilla very infrequently because of it (that's a good thing).  The problem to solve is figuring out when (and when not) to comment and change a bugzilla.  This is only exposed because of people pulling hundreds of extra commits into their PRs, which goes against our workflow anyway.

jrafanie:
I forget how this works but can't we limit most PR checks to X number of commits?

bdunne:
image

carbonin:
We probably don't have any visibility into the number of commits because this was a commit checker (rather than a commit_range checker)

I had another fix to just change this to check the current state of the BZ and not comment if it was > ON_DEV

Fryguy:
Closing for now since I merged #194

PR #194, opened 2016-06-20
Check the bug status before making a comment

carbonin:
Sometimes people make PRs with bad rebases that include many commits which should not have been in the PR. This has caused the BugzillaPrChecher to comment on every BZ referenced in those commits with this new PR link.

This change will not completely solve the issue, but will at least prevent the bot from adding comments to BZs which obviously already have fixes (statuses not NEW, ASSIGNED, ON_DEV) and should not be being referenced in PR commits.

Alternative to #193 

@bdunne what do you think?


bdunne:
@carbonin This safety check is probably a good idea.

In theory we shouldn't be making changes to anything in POST, MODIFIED ON_QA or VERIFIED, but should we still allow comments there?  Might happen in the case of a backport.

PR #196, opened 2016-06-22
Upgrade awesome_spawn to v1.4.1

Fryguy:
This version fixes a bug with Pathname objects, which is required for
the bot to work correctly when cloning from GitHub.


bdunne:
:+1:

PR #201, opened 2016-07-01
Remove Gemfile.lock and lock down specific gems in Gemfile

Fryguy:
The only real reason we had a Gemfile.lock was to lockdown specific
gems, but since the Gemfile.lock is resolved on Linux, a few other
dependencies don't work on other systems like MacOSX.

@bdunne Please review


bdunne:
:+1: LGTM.  Will merge when green

PR #203, opened 2016-07-05
Extract GemChangesLabeler from DiffFilenameChecker

Fryguy:
This removes the batch commenting for gem changes.  For those parties
involved, this was not a useful feature anyway, so this simplifies this
down to just labeling.

@bdunne Please review

cc @JPrause @simaishi ... just letting you know I am removing the PR comments about gemfile changes.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -7,8 +7,7 @@ github_credentials:
   password:
 commit_monitor:
   bugzilla_product:
-gemfile_checker:
-  pr_contacts: []
+gem_changes_labeler:
   enabled_repos: []
 github_notification_monitor:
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -7,8 +7,7 @@ github_credentials:
   password:
 commit_monitor:
   bugzilla_product:
-gemfile_checker:
-  pr_contacts: []
+gem_changes_labeler:
```
Fryguy: @bdunne Note this change...we will have to update the deploted settings.local.yml.

bdunne:
:+1: LGTM

PR #207, opened 2016-07-05
Add a diff content checker to match Strings, Symbols and Regexps

### LINKED ISSUE #32: Complain in PR's when illegal values are found in the diff content
The bot should complain in a PR if the diff contents contain certain unwanted strings like "cfme".

bdunne:
This allows the bot to comment when an offender is included in a PR diff.
i.e. Someone leaves a `puts` in their code, however it can be ignored in the tools directory where it is expected.

Fixes #32

### DIFF
--- a/app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
+++ b/app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
@@ -1,0 +1,45 @@
+module CommitMonitorHandlers::Batch
+  class GithubPrCommenter::DiffContentChecker
+    include Sidekiq::Worker
+    sidekiq_options :queue => :miq_bot
+
+    include BatchEntryWorkerMixin
+    include BranchWorkerMixin
+
+    def perform(batch_entry_id, branch_id, _new_commits)
+      return unless find_batch_entry(batch_entry_id)
+      return unless find_branch(branch_id, :pr)
+      complete_batch_entry({:result => process_lines})
+    end
+
+    private
+
+    def process_lines
+      @offenses = []
+
+      branch.git_service.diff.with_each_line do |line, _parent_hunk, parent_patch|
+        next unless line.addition?
+        check_line(line, parent_patch)
+      end
+
+      @offenses
+    end
+
+    def check_line(line, patch)
+      file_path = patch.delta.new_file[:path]
+      Settings.diff_content_checker.each do |offender, options|
+        next if options.except.try(:any?) { |except| file_path.start_with?(except) }
+
+        regexp = options.type == :regexp ? Regexp.new(offender.to_s) : /\b#{Regexp.escape(offender.to_s)}\b/i
+        add_offense(offender, options, file_path, line) if regexp.match(line.content)
+      end
+    end
+
+    def add_offense(offender, options, file_path, line)
+      line_number = line.new_lineno
+      message     = options.message || "Detected `#{offender}`"
+
+      @offenses << OffenseMessage::Entry.new(options.severity, message, file_path, line_number)
+    end
+  end
+end
--- a/spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker_spec.rb
+++ b/spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker_spec.rb
@@ -1,0 +1,76 @@
+require 'spec_helper'
+
+describe CommitMonitorHandlers::Batch::GithubPrCommenter::DiffContentChecker do
+  let(:batch_entry)        { BatchEntry.create!(:job => BatchJob.create!) }
+  let(:branch)             { create(:pr_branch) }
+  let(:content_1)          { "def a(variable)" }
+  let(:content_2)          { "puts 'hi'" }
+  let(:file_path)          { "tools/abc.rb" }
+  let(:git_service_double) { double("GitService", :diff => diff) }
+
+  let(:diff) do
+    rugged_delta  = double("RuggedDelta", :new_file => {:path => file_path})
+    rugged_line_1 = double("RuggedLine",  :addition? => true, :content => content_1, :new_lineno => 1)
+    rugged_line_2 = double("RuggedLine",  :addition? => true, :content => content_2, :new_lineno => 2)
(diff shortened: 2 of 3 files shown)


app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def check_line(line, patch)
+      Settings.diff_content_checker.each do |offender, options|
+        next if options.except.try(:any?) { |except| patch.delta.new_file[:path].start_with?(except) }
+
+        if options.type == :regexp
+          add_offense(offender, options, patch, line) if Regexp.new(offender.to_s).match(line.content)
+        else
+          add_offense(offender, options, patch, line) if line.content.downcase.include?(offender.to_s)
```
Fryguy: This can be simplified, but it's ok for now...

```ruby
regexp = options.type == :regexp ? Regexp.new(offender.to_s) : /#{Regexp.escape(offender)}/i
add_offense(offender, options, patch, line) if regexp.match(line.content)
```

app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+      end
+    end
+
+    def add_offense(offender, options, patch, line)
+      file_path   = patch.delta.new_file[:path]
```
Fryguy: This file path is duplicated with line 30...probably could extract up above and then only pass the path around instead of the whole patch.

spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  let(:rugged_delta)       { double("RuggedDelta", :new_file => {:path => file_path}) }
+  let(:rugged_patch_1)     { double("RuggedPatch", :hunks => [rugged_hunk_1], :delta => rugged_delta) }
+  let(:rugged_hunk_1)      { double("RuggedHunk", :lines => [rugged_line_1, rugged_line_2]) }
+  let(:rugged_line_1)      { double("RuggedLine", :addition? => true, :content => content_1, :new_lineno => 1) }
+  let(:rugged_line_2)      { double("RuggedLine", :addition? => true, :content => content_2, :new_lineno => 2) }
+
+  before do
+    stub_sidekiq_logger
+    stub_job_completion
+    github_service
```
Fryguy: I don't think you need the github service in this file.

spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'spec_helper'
+
+describe CommitMonitorHandlers::Batch::GithubPrCommenter::DiffContentChecker do
+  let(:batch_entry)        { BatchEntry.create!(:job => BatchJob.create!) }
+  let(:branch)             { create(:pr_branch) }
+  let(:diff)               { GitService::Diff.new(rugged_diff) }
+  let(:git_service_double) { double("GitService", :diff => diff) }
```
Fryguy: Can you copy the pattern from my other PR for this? I thought it came out a little cleaner because 1) no need to call it a double in the let name and 2) You don't need the diff object anywhere in the spec, so might as well hide it right in the double creation.

app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
Fryguy: This seems a little too relaxed...If I have a word checker looking for the word `puts`, this will flag on `inputs`.  Perhaps instead for the string case, you should use word boundaries with `/\b#{Regexp.escape(offender)}\b/i`.

spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker_spec.rb
Fryguy: This is what I mean: https://github.com/ManageIQ/miq_bot/blob/master/spec/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_filename_checker_spec.rb#L4

Trying to avoid creating a ton of lets for things that are essentially "private" values that aren't used by specs directly.  I might even consider collapsing that entire rugged_diff definition into:

```ruby
let(:diff) do
  rugged_line_1  = double("RuggedLine", :addition? => true, :content => content_1, :new_lineno => 1) }
  rugged_line_2  = double("RuggedLine", :addition? => true, :content => content_2, :new_lineno => 2) }
  rugged_hunk_1  = double("RuggedHunk", :lines => [rugged_line_1, rugged_line_2]) }
  rugged_delta   = double("RuggedDelta", :new_file => {:path => file_path}) }
  rugged_patch_1 = double("RuggedPatch", :hunks => [rugged_hunk_1], :delta => rugged_delta) }
  rugged_diff    = double("RuggedDiff", :patches => [rugged_patch_1]) }
  GitService::Diff.new(rugged_diff)
end

File: app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb [hunk 5]
```diff
@@ -0,0 +1,48 @@
+module CommitMonitorHandlers::Batch
+  class GithubPrCommenter::DiffContentChecker
+    include Sidekiq::Worker
+    sidekiq_options :queue => :miq_bot_glacial
```
Fryguy: Should this be on the regular queue?

app/workers/commit_monitor_handlers/batch/github_pr_commenter/diff_content_checker.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    include BatchEntryWorkerMixin
+    include BranchWorkerMixin
+
+    def perform(batch_entry_id, branch_id, _new_commits)
+      return unless find_batch_entry(batch_entry_id)
+      return unless find_branch(branch_id, :pr)
```
Fryguy: We might want to consider enable/disable for this feature by repo...not sure the best way to break that up or if we should worry about it right now.

PR #208, opened 2016-07-06
Fix "undefined method html_url for nil"

bdunne:
Sometimes the repo is nil, so we need a different html_url.
i.e. User creates a PR then deletes their fork


bdunne:
@Fryguy Please review

PR #209, opened 2016-07-06
Diff content checker followup

bdunne:
- Accidentally deleted method
- default value in settings.yml


bdunne:
@Fryguy Please review

PR #210, opened 2016-07-06
Ensure 'lib' is eager loaded to avoid missing constants in workers

bdunne:
Make TravisEvent::Listener a library, add a rake task to run it


bdunne:
@Fryguy Please Review

PR #211, opened 2016-07-08
Loosen .ruby-version to any 2.2

Fryguy:
This file drives me nuts because I don't have 2.2.3 locally, so anytime I do cd or git checkout I lose my Ruby.  This PR loosens to any 2.2 series Ruby.


chrisarcand:
@Fryguy Sadly this change isn't compliant with all Ruby managers (rbenv purposely doesn't allow this, annoyingly). If it [understandably] bothers you that it locks a version, mind if I just remove it?

`rbenv: version `2.2' is not installed`

chrisarcand:
IIRC I locked it to 2.2.3 because that is what the box runs, but...meh.

bdunne:
@chrisarcand I think this should go since it doesn't represent the version that the deployed machine is running.  Do you want to make a PR?

chrisarcand:
Will do, I regret adding it in the first place. <3

PR #212, opened 2016-07-09
Add sql migration labeler

Fryguy:
This is mostly copied from the gem changes labeler.  There is a bit of a pattern, but I'm not ready to DRY it up into a separate class until I see a 3rd one ([Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and all that)

@bdunne  Please review.


bdunne:
:+1: LGTM

PR #213, opened 2016-07-11
Denormalize PullRequestMonitor

Fryguy:
Remove extra classes and simplify where the responsibilities lie
- Repo model has the knowledge to create/prune PR branches
- PullRequestMonitor deals with talking to GitHub and then calling those
  methods


app/workers/pull_request_monitor.rb
```diff
@@ -10,12 +10,28 @@ def perform
...
     else
-      Repo.includes(:branches).each do |repo|
-        # TODO: Need a better check for repos that *can* have PRs
-        next unless repo.upstream_user
+      Repo.includes(:branches).each { |repo| process_repo(repo) }
+    end
+  end
 
-        RepoProcessor.process(repo)
+  def process_repo(repo)
```
chessbyte: This feels like it should be an instance method in the `Repo` class.

Fryguy: This is the guts of the purpose of the PullRequestMonitor.  Otherwise you'd need a method on the Repo model called `process_for_pull_request_monitor`, which feels wrong.

Fryguy:
I don't like that the Repo model itself has GitHub knowledge...I want to remove that too (especially as it causes the PR Monitor to hit GH twice), but felt that was took much for this PR.

app/models/repo.rb
```diff
@@ -80,23 +85,48 @@ def pr_branch_names
...
-    pr_branch_names - github_branch_names
+  # @param expected [Array<Hash>] The desired state of the PR branches.
+  #   Caller should pass an Array of Hashes that contain the PR's number,
+  #   html_url, and merge_target.
+  def synchronize_pr_branches(expected)
+    raise "repo cannot not have PR branches" unless can_have_prs?
+
+    git_fetch # TODO: Let's get rid of this!
+
+    transaction do
```
bdunne: Do we need to lock the branch records for the duration of the transaction?  I'm thinking other workers could come along and change something while we're in a transaction.

bdunne:
Overall looks good

PR #215, opened 2016-07-13
Don't put effort into unmergeable PRs

bdunne:
If a PR is not mergeable, trying to look at the diff will raise a Rugged::IndexError and put the item into the retry queue


bdunne:
@Fryguy Please review

PR #219, opened 2016-07-20
Add PullRequestMonitorHandlers::MergeTargetTitler

### LINKED ISSUE #174: PRs made to non-master branch should be labeled
We can use labels, or we can modify the PR title to say the branch name.

Fryguy:
which will change a PR title to tag it with the target branch if it is
not master.

Example:
  PR titled "My PR title" is created targeting branch "foo", so PR title
  will be changed to "[FOO] My PR title"

Closes #174

In order to support this new type of handler, this PR adds reacting to PR changes and calling dedicated handlers.

@bdunne Please review.

### DIFF
--- a/app/workers/pull_request_monitor_handlers/merge_target_titler.rb
+++ b/app/workers/pull_request_monitor_handlers/merge_target_titler.rb
@@ -1,0 +1,39 @@
+class PullRequestMonitorHandlers::MergeTargetTitler
+  include Sidekiq::Worker
+  sidekiq_options :queue => :miq_bot
+
+  include BranchWorkerMixin
+
+  def perform(branch_id)
+    return unless find_branch(branch_id, :pr)
+    return unless verify_branch_enabled
+
+    process_branch
+  end
+
+  private
+
+  def process_branch
+    apply_title if merge_target != "master" && !already_titled?
+  end
+
+  def already_titled?
+    pr_title_tags.map(&:downcase).include?(merge_target.downcase)
+  end
+
+  def apply_title
+    branch.repo.with_github_service do |github|
+      logger.info("Updating PR #{pr_number} with title change for merge target #{merge_target}.")
+      github.pull_requests.update(
+        :user   => github.user,
+        :repo   => github.repo,
+        :number => pr_number,
+        :title  => new_pr_title
+      )
+    end
+  end
+
+  def new_pr_title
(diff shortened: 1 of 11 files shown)


Fryguy:
cc @chessbyte ...once this is merged and deployed, you will not need to manually mark PRs as `[DARGA]`

app/workers/pull_request_monitor_handlers/merge_target_titler.rb
```diff
@@ -0,0 +1,39 @@
...
+        :user   => github.user,
+        :repo   => github.repo,
+        :number => pr_number,
+        :title  => new_pr_title
+      )
+    end
+  end
+
+  def new_pr_title
+    "[#{merge_target.upcase}] #{pr_title}"
```
bdunne: Do you need to truncate the string at a max length?

spec/workers/pull_request_monitor_handlers/merge_target_titler_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    stub_settings(:merge_target_titler => {:enabled_repos => [branch.repo.name]})
+  end
+
+  context "when the branch has a non-master merge target" do
+    before do
+      branch.update_attributes!(:merge_target => "darga")
+    end
+
+    it "and not already titled" do
+      expect(github_service).to receive(:user).and_return("SomeUser")
```
bdunne: wasn't expecting this...

app/workers/pull_request_monitor_handlers/merge_target_titler.rb
bdunne: Do we need to worry about other title workers changing the title at the same time?

Fryguy: I'll check on the length...good question on the concurrency aspect.  I may need a helper method like add_pr_title_tag that is wrapped in a mutex.

spec/workers/pull_request_monitor_handlers/merge_target_titler_spec.rb
Fryguy: Yeah, for some reason the github api requires that I "resend" the user and repo...it's not being honored from the higher level calls.  It feels like a bug in the gem.

Issue #34, opened 2015-02-13
Auto-label (or unlabel) WIP PRs


Fryguy:
Can be done simply at first by checking the PR title `.start_with?("WIP") || .start_with?("[WIP]")`

Fryguy:
Completed in #220

PR #220, opened 2016-07-20
Add PullRequestMonitorHandlers::WipLabeler

Fryguy:
This is the first part of supporting automatic WIP labeling and handles
a user changing the title to have [WIP].  Not yet handled is when a user
removes [WIP], nor if they manually add the label with an miq-bot command

Issue #34

~~Built on top of #219 since that one has the basis for PR handlers.  See only the last commit.~~ #219 is merged and this has been rebased.

@bdunne Please review

### DIFF
--- a/app/models/git_hub_api/issue.rb
+++ b/app/models/git_hub_api/issue.rb
@@ -1,5 +1,7 @@
 module GitHubApi
   class Issue
+    WIP_REGEX = /^(?:\s*\[wip\])+/i
+
     attr_accessor :comments, :number, :body, :author, :created_at
 
     def initialize(octokit_issue, repo)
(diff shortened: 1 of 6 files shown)


Fryguy:
cc @chessbyte ...once this is merged and deployed, you will not need to manually mark PRs with the wip label if the user changes the title to add `[WIP]` (case insensitive, so `[wip]` or `[Wip]` work too).  However, this does not handle the following cases (TBD in follow up PRs)

- If a user removes `[WIP]` from the title, it will not remove the label
- If a user says `@miq-bot add_label wip`, it will not add `[WIP]` to the title
- If a user says `@miq-bot remove_label wip`, it will not remove `[WIP]` from the title...interestingly, it *may* put the label back if `[WIP]` is still in the title.

Fryguy:
@chrisarcand I think this PR might be good to go now that @jjlangholtz has updated it to fill out the rest of the bullet points.

chrisarcand:
Yes I think it is too.

It's unfortunate that this includes code that utilizes both Github gems 😕  It's also unfortunate that I actually just finished deleting the entire GitHubApi module. However, we can just merge and deploy and I'll just deal with it. Note that it means once I rebase this code will look quite a bit different when you see it next. Thanks @jjlangholtz !

PR #221, opened 2016-07-22
[WIP] Linters - eslint, jscs, jshint & scss

himdel:
This adds two more linters to miq-bot - eslint (for javascript) and scss-lint (for sass/scss).

Their JSON output has a different structure, so this also adds a `convert_parsed` method to Linter::Base, and overrides to convert to the rubocop format:

eslint output:

```
[
  {
    filePath: "foo/bar",
    messages: [
      {
        ruleId: "baz",
        severity: 1,
        message: "quux",
        line: 4,
        column: 5,
      },
    ],
  },
]
```

scss output:

```
{
  "foo/bar": [
    {
      line: 4,
      column: 5,
      severity: "warning",
      reason: "quux",
      linter: "baz",
    },
  ],
}
```

converted output:

```
{
  files: [
    {
      path: "foo/bar",
      offenses: [
        {
          severity: "warning",
          message: "quux",
          location: {
            line: 4,
            column: 5,
          },
          cop_name: "baz",
        },
      ],
    },
  ],
  summary: {
    target_file_count: 1,
    offense_count: 1,
  },
}
```

TODO:
- [ ] test converters
- [ ] requires globally installed eslint-cli in the VM
- maybe add stylelint support, ManageIQ/manageiq-ui-classic#2541

### DIFF
--- a/lib/linter/base.rb
+++ b/lib/linter/base.rb
@@ -27,13 +27,7 @@ module Linter
         end
       end
 
-      begin
-        offenses = JSON.parse(result.output.chomp)
-      rescue JSON::ParserError => error
-        logger.error("#{log_header} #{error.message}")
-        logger.error("#{log_header} Failed to parse JSON result #{result.output.inspect}")
-        return failed_linter_offenses("error parsing JSON result")
-      end
+      offenses = parse_output(result.output)
       logger.info("#{log_header} Completed run with offenses #{offenses.inspect}")
       offenses
     end
@@ -67,6 +61,21 @@ module Linter
 
     def files_to_lint
       @files_to_lint ||= filtered_files(diff_service.new_files)
+    end
+
+    def parse_output(str)
+      begin
+        convert_parsed(JSON.parse(str.chomp))
+      rescue JSON::ParserError => error
+        logger.error("#{log_header} #{error.message}")
+        logger.error("#{log_header} Failed to parse JSON result #{result.output.inspect}")
+        return failed_linter_offenses("error parsing JSON result")
+      end
+    end
+
+    # overriden in linters with different JSON format
+    def convert_parsed(hash_or_array)
+      hash_or_array
     end
 
     def run_linter(dir)
--- a/lib/linter/eslint.rb
+++ b/lib/linter/eslint.rb
@@ -19,5 +19,27 @@ module Linter
         file.end_with? '.js'
       end
     end
+
+    def convert_parsed(array)
+      files = array.map { |f| convert_file(f) }
+
+      {:files   => files,
+       :summary => {:offense_count     => files.reduce(0) { |memo, f| memo + f.offenses.count },
+                    :target_file_count => files.count}}
+    end
+
+    def convert_file(file)
+      {:path     => file[:filePath],
+       :offenses => file[:messages].map { |m| convert_offense(m) }}
+    end
+
+    SEVERITY = ['off', 'warning', 'error']
+    def convert_offense(offense)
+      {:severity => SEVERITY[offense[:severity]],
+       :message  => offense[:message],
+       :location => {:line   => offense[:line],
+                     :column => offense[:column]},
+       :cop_name => offense[:ruleId]}
+    end
   end
 end
--- a/lib/linter/eslint.rb
+++ b/lib/linter/eslint.rb
@@ -1,0 +1,23 @@
+module Linter
+  class ESLint < Base
+    private
+
+    def config_files
+      [".eslintrc.js", ".eslintrc.json"]
+    end
+
+    def linter_executable
+      'eslint .'
+    end
+
+    def options
+      {:format => 'json'}
+    end
+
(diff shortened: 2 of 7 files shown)


romanblanco:
related to issue https://github.com/ManageIQ/manageiq/issues/2389

Fryguy:
You may also need to deal with this...perhaps not this PR, but... https://github.com/ManageIQ/miq_bot/blob/a056a9bf8f8e5872e39db6cd513c8d7af5384109/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb#L45

Fryguy:
@himdel Is this good to go?  Or is there more stuff to be done?

himdel:
@Fryguy this it's really missing only 2 things now..

a) verify that all the linter's JSON output formats are compatible enough 

b) cache node_modules - in the end I think the festest solution is to shasum package.json and .. essentially set `NODE_PATH=whatever$(sha1sum packages.json)`

chessbyte:
@himdel please rebase and push again

lib/linter/base.rb
```diff
@@ -69,6 +63,21 @@ def files_to_lint
...
+      begin
+        convert_parsed(JSON.parse(str.chomp))
+      rescue JSON::ParserError => error
+        logger.error("#{log_header} #{error.message}")
+        logger.error("#{log_header} Failed to parse JSON result #{result.output.inspect}")
+        return failed_linter_offenses("error parsing JSON result")
+      end
+    end
+
+    # overriden in linters with different JSON format
```
Fryguy: comment typo `overriden` -> `overridden`

lib/linter/eslint.rb
```diff
@@ -0,0 +1,45 @@
...
+      files.select do |file|
+        file.end_with? '.js'
+      end
+    end
+
+    def convert_parsed(array)
+      files = array.map { |f| convert_file(f) }
+
+      {:files   => files,
+       :summary => {:offense_count     => files.reduce(0) { |memo, f| memo + f.offenses.count },
```
Fryguy: I believe `files.reduce(0) { |memo, f| memo + f.offenses.count }` can be simplified to `files.map { |f| f.offenses.count }.reduce(:+)`

Fryguy:
#406 will likely supercede this

himdel:
Or https://github.com/ManageIQ/miq_bot/pull/500 will :)

Fryguy:
going to close this in favor of Pronto work as it's being revived.

himdel:
@NickLaMuro for eslint, we need the following npm packages:

* `babel-eslint`
* `eslint`
* `eslint-config-airbnb`
* `eslint-config-angular`
* `eslint-plugin-import`
* `eslint-plugin-jest`
* `eslint-plugin-jsx-a11y`
* `eslint-plugin-react`

(you can assume current versions, I'll be updating them on ui-classic side soon; eslint-{config,plugin}-standard eslint-config-promise are unused)

I guess it makes sense to just copy those dependencies to manageiq-style, and update both places.

---

One more catch .. the `.eslintrc.json` configfile can be in any directory, we need them all (including the relative symlinks, but they can be copies).
(It inherits from parent dirs, so we can have spec-only globals, etc.)

NickLaMuro:
Turns out it is probably for the best that we aren't using `proto` for this as it looks like the gem under the hood it uses is super ancient, and uses a version of `eslint` that is very much out of date:

https://github.com/zendesk/eslintrb/tree/master/vendor

And probably doesn't have all of he plugins you want.

* * *

That said, the list of plugins is going to be a bit of a nightmare to source, so we might want to consider vendoring them into this repo to avoid inconsistencies.  This would mean that when run via a potential future "cli client", there is no concern about potentially getting the wrong gems.

Alternatively, we could consider creating a stand-alone repo for doing just that, and it can also have a `gem` component for ruby projects that want to consume it, with the process of updating it just being done with whatever standard JS means is preferred.

himdel:
Oh, well, a CLI client would be great for development, but that may just be complicating matters now.

What about something as simple as having a manageiq-style package.json with the dependencies (and possibly a .nvmrc with a node version), and a base .eslintrc.json to apply when no root eslintrc.json is present.

The primary goal is to get this running on the bot, and it can just be statically installed there.

himdel:
> The primary goal is to get this running on the bot, and it can just be statically installed there.

In fact, I'd propose this directory structure, if doable...

```
root/
  package.json (from manageiq-style)
  .eslintrc.json (from manageiq-style)
  node_modules/ (by running yarn in root/)
  $repo/
```

That way, `yarn run eslint $repo` should work, pick up `$repo/.eslintrc.json` if present, fallback to `root/.eslintrc.json`  if not.
And `root/` can be cached on the bot, only updated when manageiq-style changes, while `root/$repo` gets regularly updated.

himdel:
(and since `$repo` can be a tempdir with a random name, it should still work in parallel)

PR #222, opened 2016-10-11
Add the .rubocop_base.yml file to the linter config files list

carbonin:
This was added in https://github.com/ManageIQ/manageiq/pull/11750 so that we could avoid inheriting from a remote repo as that is not supported by code climate.

@bdunne @jrafanie please review


jrafanie:
👍 Nice @carbonin

PR #224, opened 2016-10-13
Removed needless requires, removes warnings

chrisarcand:
These are auto loaded by Rails and the re-require causes warnings of
redefining the BugNotFoundError therein when you load the bot.


chrisarcand:
@miq-bot assign @Fryguy

PR #232, opened 2016-11-02
Fix issue where blob_at would extract from the merge and not the tip.

Fryguy:


lib/git_service/branch.rb
```diff
@@ -8,10 +8,11 @@ def initialize(branch)
     end
 
     def blob_at(path) # Rugged::Blob object for a given file path on this branch
-      blob_data = merge_tree.path(path)
+      blob_data = tip_tree.path(path)
       blob = Rugged::Blob.lookup(rugged_repo, blob_data[:oid])
       (blob.type == :blob) ? blob : nil
     rescue Rugged::TreeError
+      nil
```
chrisarcand: Is this just making something explicit that I thought was already implicit?

jrafanie: Good point.  @Fryguy can you reword the cleanup commit to explain why the two changes.  You mentioned the GC.start move was because other places can hit it and the `merge_index` was really the problem.  But that was in the office, it's not clear in the commit message.

Fryguy: Ok done.

jrafanie:
Looks good.

PR #233, opened 2016-12-06
Rescue Github::Error::NotFound during remove_label if PR doesn't have that label

bdunne:


bdunne:
@Fryguy Please review

PR #243, opened 2017-01-06
Remove and clarify the HTTPS git URL restriction

Fryguy:
@bdunne Please review.

This was added because it was assumed that SSH URLs wouldn't work, but they work fine as long as the bot account has the keys in place.


bdunne:
Like the idea, just needs a spec update.  Don't remember any other reasons for this limit.

Fryguy:
DOH! on it.

Fryguy:
@bdunne Updated and also beefed up the spec while I was in there.

PR #256, opened 2017-01-16
Bump rubocop to 0.46.0 and haml-lint to 0.19.0

Fryguy:
@bdunne Please review

I also updated the various test to re-record the json files, which uncovered a few very minor issues.


Fryguy:
Note that rubocop 0.47.0 is actually out, but I can't upgrade to it yet because it breaks haml-lint (known issue that was opened on haml-lint).

PR #264, opened 2017-01-20
Add support for Github rate limit logging

Fryguy:
~~Depends on https://github.com/ManageIQ/miq_tools_services/pull/24~~ Obviated by #265 

@bdunne @chrisarcand Please review

### DIFF
--- a/config/initializers/services_initializer.rb
+++ b/config/initializers/services_initializer.rb
@@ -1,6 +1,7 @@
 unless Rails.env.test?
   BugzillaService.credentials = Settings.bugzilla_credentials
   GithubService.credentials   = Settings.github_credentials
+  PivotalService.credentials  = Settings.pivotal_credentials
 
-  PivotalService.credentials = Settings.pivotal_credentials
+  GithubService.logger = Sidekiq::Logging.logger
 end


config/initializers/services_initializer.rb
```diff
@@ -1,6 +1,7 @@
 unless Rails.env.test?
   BugzillaService.credentials = Settings.bugzilla_credentials
   GithubService.credentials   = Settings.github_credentials
+  PivotalService.credentials  = Settings.pivotal_credentials
 
-  PivotalService.credentials = Settings.pivotal_credentials
+  GithubService.logger = Sidekiq::Logging.logger
```
chrisarcand: I don't think I understand this line 😕 What's the intent here?

Fryguy: The intent is that the GithubService will log into the Sidekiq logger.  The only logging we have enabled for the GithubService is the RatelimitLogger, but we could add others in the future (like request/response logging).

chrisarcand: Ah I suppose it makes sense to add rate limit logging to the sidekiq log wherein the workers who are doing the calling are doing the work :D derp.

PR #268, opened 2017-01-23
Fix issue where existing title tags may not have space delimiters.

Fryguy:
@chrisarcand Please review.  This should fix the [DARGA][DARGA] issue.

### DIFF
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -63,7 +63,7 @@ class Branch < ActiveRecord::Base
   end
 
   def pr_title_tags
-    pr_title.to_s.match(/^(?:\s*\[\w+\])+/).to_s.split.map { |s| s[1...-1] }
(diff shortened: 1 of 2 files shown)


app/models/branch.rb
```diff
@@ -63,7 +63,7 @@ def pr_number
   end
 
   def pr_title_tags
-    pr_title.to_s.match(/^(?:\s*\[\w+\])+/).to_s.split.map { |s| s[1...-1] }
+    pr_title.to_s.match(/^(?:\s*\[\w+\])+/).to_s.gsub("[", " [").split.map { |s| s[1...-1] }
```
chrisarcand: This will add a double space on ones that have a space [but it doesn't matter as the regex accounts for any amount of whitespace], right?

Fryguy: yeah, it ends up not mattering because .split handles multiple consecutive spaces.  In fact, the other tests prove this as well.

chrisarcand: Yup tests prove everything's great, just asking for my own understanding 💙  Looks great

PR #270, opened 2017-01-25
Add observing of Github rate limit via InfluxDB

chrisarcand:
So the plan with this is that I'm building a simple graph page in the UI to
see trends in Github usage, complete with tagging of the current Bot SHA
to measure actual improvments in our Github interface, caching, etc.

Sadly my Javascript is less than stellar and it's taking me just a bit longer
than I'd like. I really want to get the bot making measurements though so we
have an initial benchmark to work with. Therefore, submitting this to get set
up and start measuring while I finish the front end portion.


lib/github_usage_tracker.rb
```diff
@@ -0,0 +1,21 @@
+class GithubUsageTracker
+  def self.record_datapoint
+    new.record_datapoint
+  end
+
+  def record_datapoint
+    # TODO Clean this mess up
```
chrisarcand: This will be cleaned up with the direction we've discussed, which I'll be doing once I get this little feature wrapped up.

lib/tasks/github_usage_tracker.rake
```diff
@@ -0,0 +1,6 @@
+namespace :github_usage_tracker do
+  desc "Record a current datapoint of GitHub API usage"
+  task :record_datapoint => :environment do
+    GithubUsageTracker.record_datapoint
+  end
```
chrisarcand: This will be run on a cron. Can add more details about setting up InfluxDB and crons in another PR.

config/initializers/influxdb-rails.rb
```diff
@@ -0,0 +1,24 @@
+InfluxDB::Rails.configure do |config|
+  config.influxdb_database = "miq_bot_#{Rails.env}"
```
bdunne: Can we pull these values from Settings instead?

lib/tasks/github_usage_tracker.rake
bdunne: Can we make it a SideKiq worker and schedule it in SideTiq instead?

config/initializers/influxdb.rb
```diff
@@ -0,0 +1,24 @@
+InfluxDB::Rails.configure do |config|
+  config.influxdb_database = "miq_bot_#{Rails.env}"
```
bdunne: Can these values be pulled from Settings instead?  I'd prefer to have as much of the config in one place as possible.

lib/tasks/github_usage_tracker.rake
chrisarcand: I thought of that but figured that placing it on a Sidekick queue displaces time...but I'm hardly knowing what that means, really. If I do that, can I get the worker to go immediately? Given the crazy abilities of the TSDB and that this is a good testing ground to see if we can do things like this, I kinda wanted to try observations only a minute a part to start.

config/initializers/influxdb.rb
chrisarcand: Sure can.

lib/tasks/github_usage_tracker.rake
chrisarcand: It doesn't seem like it: https://github.com/mperham/sidekiq/wiki/Scheduled-Jobs#checking-for-new-jobs

bdunne: If we put it on the regular queue it could get backed up behind other work, but I would be fine with adding another queue for this kind of stuff.  Once it's on the queue, it should be picked up immediately.
```ruby
class MyWorker
  include Sidekiq::Worker
  include Sidetiq::Schedulable

  recurrence { minutely }
end
```

bdunne: We do this for a couple workers already.  See https://github.com/ManageIQ/miq_bot/blob/master/app/workers/batch_job_monitor.rb#L8

chrisarcand: Ok; I'm good with it if it can have a queue separate from the current ones. I just wanted to make sure it has a higher degree of likelihood that it wouldn't get stuck or displaced due to all the other things we queue up but this seems good. 👍

chrisarcand: We can also consolidate it back in to a normal queue with less frequency later if we want; for our current priorities I'd like to keep a fair number of as-accurate-as-possible observations.

chrisarcand:
With the given feedback and a couple other developments on the UI side, I'm closing this and reopening a new one soon.

PR #271, opened 2017-01-26
GitHub API Usage feature

chrisarcand:
Introduces a new UI for viewing GitHub rate limit measurements, taken every minute and stored in InfluxDB, a fun little time series database.

The query determining the graph data are measurements over the last 12 hours grouped by the average rate limit over 1 minute intervals. That's a little over 700 datapoints. I made a sample dataset in my local instance with over 37,000 datapoints on the rate_limit measurement.

Some notes:

* The current hardcoded query in `rate_limit_measurements` is the most helpful graph I could come up with for now. TBH I think the next step is just to go straight to Grafana (as discussed with @Fryguy) because it would be honestly less time and more helpful in the long run than designing a bunch of different charts in Google Charts. In short, "This is good enough to be helpful for now and let's circle back and make it awesome later."
* Although it's not used here, every datapoint is tagged with the current code's SHA. Later on, we can come up with some neat comparisons of different versions of the bot and the improvements that we (hopefully) made.
* I've put a 3 month retention policy on rate limit datapoints. That might be too much, I don't know. I just figured we should *something* from keeping the bot overflowing with the number of datapoints that we'll have. Note I added some stuff to the README to set that up manually; that's because I don't think this is worth coming up with some sort of formal 'running Influx migrations' implementation yet.
* I found a bug in the Influx client that made me do that ugly `#log` method hack. 

Replaces #270


chrisarcand:
Ugh, I tried generating some realistic sample data in one minute increments and the graph is somewhat borked; the precision might be broken in the Ruby client. Lemme glance at this a bit more.

Gemfile
```diff
@@ -6,6 +6,9 @@ gem 'rails', '~> 4.2.4'
 # Use PostgreSQL as the database for Active Record
 gem 'pg'
 
+# InfluxDB for Github rate limit tracking
+gem 'influxdb-rails'
```
bdunne: Usually we lock down the version a little bit, but this is only `0.1.11`.  Should we lock down to `~> 0.1.11`?

chrisarcand: For sure, good call.

chrisarcand:
@bdunne I chatted with @Fryguy this morning; I'm going to try and change this to instead hook in to the middleware for all Github calls and record based on the returned header, then group data by a sane time interval for the front end here. Then we won't be adding another worker or even another [free] call to Github for this.

bdunne:
ok, sounds good.

chrisarcand:
@Fryguy @bdunne Ok! I think this is finally ready for a real review. I've updated the description entirely, so look back on that.

app/views/layouts/_footer.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="page-footer">
-	<p class="text-muted">&copy; 2014 Red Hat, Inc.</p>
+  <p class="text-muted">&copy; <%= Time.now.year %> Red Hat, Inc.</p>
```
Fryguy: 😮

lib/github_usage_tracker.rb
```diff
@@ -0,0 +1,48 @@
...
+        :timestamp => timestamp ? timestamp.to_i : Time.now.to_i },
+      nil, # Allow config to determine precision
+      'twelve_weeks' # Retention policy
+    )
+  end
+
+  private
+
+  def current_bot_sha
+    @current_git_sha ||= `git rev-parse --short --verify HEAD`.strip
```
Fryguy: Can we move this over to https://github.com/ManageIQ/miq_bot/blob/master/lib/miq_bot.rb, so that it can live with it's cousin method?

Fryguy:
LGTM...I have a minor comment above but we can always change it later.

lib/github_usage_tracker.rb
chrisarcand: Oh sure, yeah that seems like a better spot.

app/models/git_hub_api.rb
```diff
@@ -14,7 +14,10 @@ def self.execute(client, cmd, *args)
     val = nil
     t = Benchmark.realtime { val = client.send(cmd, *args) }
     limit_after = client.rate_limit.remaining
-    logger.info("Executed #{cmd} #{args.inspect}...api calls remaining #{limit_after} (in #{"%0.3f" % t}s using #{limit_before - limit_after} calls)")
+    logger.info("Executed #{cmd} #{args.inspect}...api calls remaining #{limit_after} " \
```
bdunne: I'm not a fan of these multi-line strings, but don't change now.

chrisarcand: The fact that it's multiline, or the continued parsing with `\`? (would you prefer `+`?) I'm not a fan of lines with an enormous amount of columns.

bdunne: The continuation with `\`.  Maybe parts could be extracted to variables above then this just assembles them?  Or maybe it should be 2 log messages?

chrisarcand: Either is fine with me :D Intent of this change was just 'I hate stoopid long lines'

PR #273, opened 2017-01-27
Use previous datapoints in usage aggregation query

chrisarcand:
When there's no github requests in a 60 second timeframe, no data is
added to Influx. Therefore, when aggregating data for display use
the previous datapoints instead of null for that interval.

Fixes:


chrisarcand:
@bdunne or @Fryguy wanna hit that big ol green button for me?

PR #274, opened 2017-01-27
Fix limit diff from bot db

chrisarcand:
The bot is currently getting changes in its git stage because the db was
created from a previous era where Rails put limits in its columns
without you specifying. This migration is a no op in the bot's
perspectice but adds these limits for development environments so the
schema is exactly the same as what the bot currently uses.


chrisarcand:
Oooooo I like that @Fryguy should add to resume. 

Schema Master™

Fryguy:
I may be the Schema Master, but apparently I'm not a Gif Master :(

chrisarcand:
Oh boo I need to cut that one down if we can't github post it

Fryguy:
Oh interesting...I didn't realize GH had file size limits

PR #278, opened 2017-01-31
Measure Github request URIs

chrisarcand:
* Request URIs are now recorded, so we can see and quantify painpoints in our API usage
* Begins the shift to pure Octokit by configuring outside of `GithubApi`
* Reuses RateLimitLogger middleware for all Octokit requests instead of doing it's own thing


chrisarcand:
@bdunne or @Fryguy

chrisarcand:
💙

PR #287, opened 2017-02-03
Version control Gemfile.lock

chrisarcand:
*Breaking News*

This is an application, and there are no reasons that I can think of
between our production and development environments that we can't check
in the Gemfile.lock to ensure parity between those environments and
better manage keeping our development environment up to date, as is the
point of Gemfile.lock.

This Gemfile.lock is the exact one used in our current production
environment and should have no observable effect when deployed at this time.

This pull request will awkwardly conflict with #286 as the lockfile will be different but no conflict detected. Merge either and I will rebase the other.


chrisarcand:
Hold on merging this, I should ensure no unseen conflict with #286

PR #288, opened 2017-02-03
Quick GitHubApi optimizations

chrisarcand:
The commits in this pull request reduce the minimum number of API calls in each
run of GithubNotificationMonitor from 3 to 1. Given the sheer number of repos
we monitor and how often this worker is run, I'm really hoping this cuts in to our API usage a fair amount; plus, it literally deletes code to do it. ✂️ 🔥


chrisarcand:
Belated @jrafanie birthday present! 🎂

app/models/git_hub_api/user.rb
```diff
@@ -1,10 +0,0 @@
-module GitHubApi
-  class User
-    attr_accessor :username, :password, :client
-
-    def find_organization(organization_name)
-      octokit_org = GitHubApi.execute(@client, :organization, organization_name)
-      @organization  = Organization.new(octokit_org, self)
-    end
-  end
-end
```
jrafanie: I approve of this change ^ 🔥

app/models/github_notification_monitor.rb
```diff
@@ -15,9 +15,7 @@ class GithubNotificationMonitor
 
   def self.build(organization_name, repo_name)
     fq_repo_name = "#{organization_name}/#{repo_name}"
-    user         = GitHubApi.connect
-    org          = user.find_organization(organization_name)
-    repo         = org.get_repository(repo_name)
+    repo         = GitHubApi::Repo.new(Octokit::Client.new, fq_repo_name)
```
chrisarcand: Yes, the only purpose of User and Organization was to pass a client (and waste queries doing it)

PR #295, opened 2017-02-07
Remove TravisEventHandlers/Listener

chrisarcand:
This has been broken for a while and doesn't really need to be used anymore. The listener itself is removed as if we want to do stuff in the future we'll probably be using webhooks anyway.


chrisarcand:
@jrafanie Please review

✂️ 🔥 😍

PR #183, opened 2016-05-25
update to ruby 2.2.4

chriskacerguis:

PR #299, opened 2017-02-09
Add .ruby_version

chrisarcand:
Because we should be developing on the version that we're running.

2.2.6 as that's what we're currently running with.


bdunne:
@chrisarcand We've been here before and have decided to revert it too.

bdunne:
https://github.com/ManageIQ/miq_bot/pull/216
https://github.com/ManageIQ/miq_bot/pull/211
https://github.com/ManageIQ/miq_bot/pull/183
https://github.com/ManageIQ/miq_bot/pull/155

chrisarcand:
Ugggh that's' right, I forgot. 

TBH I'm of the opinion that locking down to patch is good and you should just get all the rubies on your machine and not care. But respecting @Fryguy's original issue (https://github.com/ManageIQ/miq_bot/pull/211), whatever, I don't want to open this can of worms and will just leave it to individual dev's setups. 🙇

PR #300, opened 2017-02-09
Ignore .ruby_version

chrisarcand:
Followup to #299

Makes it easy to set ruby version locally to whatever you want without gitignoring yourself, and more importantly serves as an indicator that we *intend* that we're not locking the Ruby version.


chrisarcand:
@bdunne

PR #301, opened 2017-02-09
New GithubService singleton wrapper

chrisarcand:
Introduces a simpler, betterer Github service class implemented with Octokit.

It uses a singleton pattern just like Octokit and most other such gems, and opts out of the service 'call' pattern we've been doing, as discussed at length. Rewriting sections to use this wrapper was stupid easy and straightfoward. Core functionality from the original GithubService was added (but renamed, to feel more natural with Octokit's like-methods).

I also found it easy to use over my previous OctokitWrappers implementation, so I'll be removing that one in favor of this, as well as removing the current GithubService and renaming this one, in a future PR.

Although low priority, I suggest that we do the other services in this pattern except for GitService. MiniGit is a special little snowflake and the thread safety must be maintained.

### DIFF
--- a/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
+++ b/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
@@ -45,7 +45,7 @@ module CommitMonitorHandlers::Batch
     def replace_batch_comments
       logger.info("Adding batch comment to PR #{pr_number}.")
 
-      NewGithubService.replace_comments(branch.repo.name, pr_number, new_comments) do |old_comment|
+      NewGithubService.replace_comments(fq_repo_name, pr_number, new_comments) do |old_comment|
         batch_comment?(old_comment)
       end
     end
--- a/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
+++ b/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
@@ -28,8 +28,6 @@ module CommitMonitorHandlers::Batch
 
     private
 
-    attr_reader :github
-
     def tag
       "<github-pr-commenter-batch />"
     end
--- a/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
+++ b/app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
@@ -22,7 +22,7 @@ module CommitMonitorHandlers::Batch
         return
       end
 
-      process
+      replace_batch_comments
       complete_batch_job
     end
 
@@ -42,17 +42,10 @@ module CommitMonitorHandlers::Batch
       "#{tag}**...continued**\n"
     end
 
-    def process
+    def replace_batch_comments
       logger.info("Adding batch comment to PR #{pr_number}.")
 
-      branch.repo.with_github_service do |github|
-        @github = github
-        replace_batch_comments
-      end
-    end
-
-    def replace_batch_comments
-      github.replace_issue_comments(pr_number, new_comments) do |old_comment|
+      NewGithubService.replace_comments(branch.repo.name, pr_number, new_comments) do |old_comment|
         batch_comment?(old_comment)
       end
     end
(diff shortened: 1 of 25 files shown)


chrisarcand:
@Fryguy Please review. Except for the "Typos" commit, going through this commit by commit might be easier on the eyes.

chrisarcand:
Ugh. Needs a little more work. 

@miq-bot add_label wip

chrisarcand:
@miq-bot remove_label wip

Hey this is actually ready now 🎉

app/workers/commit_monitor_handlers/batch/github_pr_commenter.rb
```diff
@@ -42,17 +42,10 @@ def continuation_header
       "#{tag}**...continued**\n"
     end
 
-    def process
+    def replace_batch_comments
       logger.info("Adding batch comment to PR #{pr_number}.")
 
-      branch.repo.with_github_service do |github|
-        @github = github
```
Fryguy: There should be an attr_reader you can kill too.

app/workers/pull_request_monitor.rb
```diff
@@ -40,15 +40,13 @@ def process_repo(repo)
   private
 
   def github_prs(repo)
-    repo.with_github_service do |github|
-      github.pull_requests.all.collect do |github_pr|
+    NewGithubService.pull_requests(repo.name).map do |github_pr|
         {
```
Fryguy: 2 space indent got funky here?

lib/new_github_service.rb
```diff
@@ -0,0 +1,77 @@
...
+  # our own use cases.
+  #
+  # You can find the official Octokit documentation at http://octokit.github.io/octokit.rb
+  #
+  # Please check the documentation first before adding helper methods, as they
+  # may already be well handled by Octokit itself.
+  #
+  class << self
+    def service
+      @service ||= \
```
Fryguy: I don't think this `\` is necessary :)

lib/new_github_service.rb
```diff
@@ -0,0 +1,77 @@
...
+      service.respond_to?(method_name, include_private)
+    end
+
+    def method_missing(method_name, *args, &block)
+      if service.respond_to?(method_name)
+        service.send(method_name, *args, &block)
+      else
+        super
+      end
+    end
```
Fryguy: This mostly duplicates the existing stuff on ServiceMixin anyway, so there may be a way to consolidate after.

spec/spec_helper.rb
```diff
@@ -42,16 +43,6 @@
...
-      MinigitService
-    ].each do |service_model|
-      allow_any_instance_of(service_model).to receive(:service).and_raise(
-        "Live execution is not allowed in specs.  Use stubs/expectations on service instead."
-      )
-    end
-  end
 end
+
+WebMock.disable_net_connect!(:allow_localhost => true)
```
Fryguy: Interesting way to solve, but only handles things that use HTTP interface. (i.e. not minigit)

chrisarcand: Oh woooops. Forgot about our special snowflake. Let's depend on this for HTTP services but I'll add minigit back for sure, as I don't think we really want to change that service interface anyway.

Fryguy:
Looks really good!

lib/new_github_service.rb
chrisarcand: Oh weird; I thought in an assignment, if you wanted the value on the line below it, you'd need this.

ie I thought to have a begin block you'd need to do this, or this:  

```ruby
@service ||= begin
  code_that_does_stuff
  more_code
  # ...
```

Neat.

spec/spec_helper.rb
chrisarcand: Fixed

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -57,17 +57,9 @@ def rubocop_comments
...
   end
 
-  def write_to_github
-    logger.info("Updating PR #{pr_number} with rubocop comment.")
-
-    branch.repo.with_github_service do |github|
-      @github = github
-      replace_rubocop_comments
-    end
-  end
```
Fryguy: @chrisarcand Also should be an attr_reader you can kill here as well (you can do in sep PR)

PR #302, opened 2017-02-09
Remove github_api dependency

chrisarcand:
**REQUIRES #301; suggest merging #301 before review to save your eyes**

:tada:

This removes the dependency on github_api as it's no longer used, and replaces the old GithubService with the new one.


Fryguy:
For reference, now that #301 is merged, reviewing https://github.com/ManageIQ/miq_bot/compare/a76c3b1...chrisarcand:github-api-removal-2

PR #303, opened 2017-02-10
Integrate OctokitWrappers in to GithubService

chrisarcand:
**REQUIRES #302. Suggest merging #302 before review for a better diff**

The new GithubService covers things much more nicely than the current
OctokitWrappers, which was based very closely on the original GitHubApi
Octokit classes.

The wrappers are still quite a useful abstraction. There's too much info
used in responses that is frequently accessed and the wrappers provide a
nice abstracted way of getting to it without caring how. So, we simplify
them a bunch and bring them under the GithubService fold. Methods that
acted like associations (like `issue.list_comments`) are removed, because
you should use the service to get that - but things that are directly
related to the response data or are custom actions (like wipifying the
title of an issue) are left alone.

We also remove a ton of Github related stuff in the Repo model and more
appropriately put it within the GithubService.

Lastly, we reimplement GithubNotificationMonitor, the only thing that was
using the OctokitWrappers, with the new interface.


chrisarcand:
@Fryguy Please review

Fryguy:
For reference, now that #302 is merged, reviewing: https://github.com/ManageIQ/miq_bot/compare/https://github.com/ManageIQ/miq_bot/pull/303/commits/9b4e06e22efdcbdf4f68b7418b700595d0dcda5e...chrisarcand:github-api-removal-3

PR #306, opened 2017-02-13
Cleanup includes for consistency

Fryguy:
@chrisarcand Please review.  This is 1 of a number of small PRs I'm going to make to cleanup various things that came from the enable/disable work.


Fryguy:
Goal here was to separate the includes to live next to their options and to make all of the workers match in structure.

PR #307, opened 2017-02-13
Add ability to skip batch entries.

Fryguy:
@chrisarcand Please review.

This adds skipping batch entries, which is important for enable/disable effort.


app/models/batch_entry.rb
```diff
@@ -12,11 +12,15 @@ def failed?
...
+  end
+
   def complete?
-    failed? || succeeded?
+    failed? || succeeded? || skipped?
   end
 
   def check_job_complete
-    job.check_complete if job && complete?
+    job.check_complete if complete? && job
```
Fryguy: This is an extremely minor performance improvement, in that it can check attributes before making a DB call.

PR #312, opened 2017-02-14
Add graphics_labeler for labeling PRs with image changes

skateman:
We would like to use this for detecting image changes in the `ui-classic` repo. FYI @epwinchell

### DIFF
--- a/spec/workers/commit_monitor_handlers/commit_range/graphics_labeler_spec.rb
+++ b/spec/workers/commit_monitor_handlers/commit_range/graphics_labeler_spec.rb
@@ -1,0 +1,30 @@
+describe CommitMonitorHandlers::CommitRange::GraphicsLabeler do
+  let(:branch)         { create(:pr_branch) }
+  let(:git_service)    { double("GitService", :diff => double("RuggedDiff", :new_files => new_files)) }
+
+  before do
+    stub_sidekiq_logger
+    stub_settings(:graphics_labeler => {:enabled_repos => [branch.repo.name]})
+    expect_any_instance_of(Branch).to receive(:git_service).and_return(git_service)
+  end
+
+  context "when there are image changes" do
+    let(:new_files) { ["image.png", "some/other/file.rb"] }
+
+    it "adds a label to the PR" do
+      expect(GithubService).to receive(:add_labels_to_an_issue).with(branch.repo.name, branch.pr_number, ["graphics"])
+
+      described_class.new.perform(branch.id, nil)
+    end
+  end
+
+  context "where there are no image changes" do
+    let(:new_files) { ["some/other/file.rb"] }
+
+    it "does not add a label to the PR" do
+      expect(GithubService).to_not receive(:add_labels_to_an_issue)
+
+      described_class.new.perform(branch.id, nil)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/workers/commit_monitor_handlers/commit_range/graphics_labeler_spec.rb
```diff
@@ -0,0 +1,30 @@
+describe CommitMonitorHandlers::CommitRange::GraphicsLabeler do
+  let(:branch)         { create(:pr_branch) }
+  let(:git_service)    { double("GitService", :diff => double("RuggedDiff", :new_files => new_files)) }
+
+  before do
+    stub_sidekiq_logger
+    stub_settings(:graphics_labeler => {:enabled_repos => [branch.repo.name]})
+    expect_any_instance_of(Branch).to receive(:git_service).and_return(git_service)
```
chrisarcand: ❌ 

`allow(branch).to receive...`

skateman: Probably it's not saving me, but I copied this [from here](https://github.com/ManageIQ/miq_bot/blob/master/spec/workers/commit_monitor_handlers/commit_range/gem_changes_labeler_spec.rb#L8) :D fixing

skateman: @chrisarcand hm, seems like it fails if i change it that way :confused:

chrisarcand: I figured ;) BUT NO THAT DOESN'T SAVE YOU 😜

chrisarcand: You need to stub https://github.com/ManageIQ/miq_bot/blob/master/app/workers/concerns/branch_worker_mixin.rb#L6-L20

To return the branch you're created here.

skateman: Will check it tomorrow, thanks!

Fryguy: Another option is just to test process_branch directly. psuedo-code idea.  This is part of the reason most of the workers have a separate process_branch method from the perform method (although I never got around to writing the tests this way).

```ruby
subject = CommitMonitorHandlers::CommitRange::GraphicsLabeler.new
subject.branch = branch
allow(branch).to receive #...
```

Fryguy:
Overall LGTM :+1:  Let's work out the specs discussion, then I am good with merging this.

Fryguy:
@chrisarcand Is #316 just a generic version of this?  Does it make sense to implement this as a rule of #316?

chrisarcand:
Oh hah, I totally forgot that this thing probably did half of what I needed to do for #316. Whoops. Yeah I think we could probably tweak it to cover this case.

chrisarcand:
Yeah @skateman I completely unknowingly just recreated a much more generalized version of this; I think I'll finish that instead as there are other purposes for this sort of thing that we want anyway. 💙

skateman:
@chrisarcand no problems, less work for me ^^ :wink:

PR #314, opened 2017-02-14
[WIP] Replace Sidetiq w/ Sidekiq-Cron

### LINKED ISSUE #282: SideTiq is no longer maintained
According the SideTiq README, SideTiq is no longer going to be maintained because it would have to be rewritten to deal with the Sidekiq 5 update.  We may want to consider alternatives (though they may also run into the same rewrite problem).  Some that I've found:

- https://github.com/moove-it/sidekiq-scheduler
- https://github.com/ondrejbartas/sidekiq-cron (So far I prefer this one)

chrisarcand:
Closes #282

This is better besides, but became more important when Sidetiq closed all of its issues and wiki, rendering most resources on how to use it completely worthless 😠 

Drops celluloid dependency, too.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,0 +1,5 @@
+schedule_file = "config/worker_schedule.yml"
+
+if File.exist?(schedule_file) && Sidekiq.server?
+  Sidekiq::Cron::Job.load_from_hash(YAML.load_file(schedule_file))
+end
--- a/Gemfile
+++ b/Gemfile
@@ -33,8 +33,8 @@ gem 'foreman', '~> 0.64.0' # v0.65.0 breaks support for the older upstart on RHE
 gem 'config'
 
 # Sidekiq specific gems
-gem 'sidekiq', '~> 4.1.1'
-gem 'sidekiq-cron', '~> 0.4.0'
+gem 'sidekiq', '~> 4.2.0'
+gem 'sidekiq-cron', '~> 0.4.5'
 gem 'sinatra', require: false
 gem 'slim'
 
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'config'
 
 # Sidekiq specific gems
 gem 'sidekiq', '~> 4.1.1'
-gem 'sidetiq', '~> 0.7.0'
+gem 'sidekiq-cron', '~> 0.4.0'
 gem 'sinatra', require: false
 gem 'slim'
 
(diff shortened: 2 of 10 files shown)


Fryguy:
This is cool.  A few comments:

- Can you add a comment to the .yml file that's human readable, so I don't have to parse cron in my head to understand it?  For example:

```yaml
batch_job_monitor:
  # minutely
  cron: '* * * * *'
  class: 'BatchJobMonitor'
  queue: 'miq_bot'

commit_monitor:
  # every 5 minutes
  cron: '*/5 * * * *'
  class: 'CommitMonitor'
  queue: 'miq_bot_glacial'
```

chrisarcand:
Added human readable comments and removed queue selection from the schedule as the workers can specify that (if I could put all sidekiq options in the schedule I'd do that but...alas, gem doesn't allow you to do that, so can't add retry boolean)

> The initializer checks if Sidekiq.server?, but we run multiple sidekiq servers.

Eh...I don't know. Unfamiliar with how we do sidekiq then...might just have to try it.

chrisarcand:
Regarding the two processes, I think we're good. Here's the check: https://github.com/mperham/sidekiq/blob/2a07a551aa8bc2356bdec736cc349c433d1408ca/lib/sidekiq.rb#L86-L88

I think the crons are only registered once. And I've tested locally with foreman running our two queues separately and it seems fine.

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,5 @@
+schedule_file = "config/worker_schedule.yml"
+
+if File.exist?(schedule_file) && Sidekiq.server?
+  Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
```
bdunne: Missing parens `Sidekiq::Cron::Job.load_from_hash(YAML.load_file(schedule_file))`

Fryguy:
The workers I'd be concerned about already check for uniqueness, so even if it does spawn multiple, I don't think it will hurt.  I think we can deploy and then watch the logs to verify

config/initializers/sidekiq.rb
Fryguy: Why didn't rubocop flag this?

chrisarcand:
@bdunne @Fryguy Let me dig just a bit further, then.

chrisarcand:
Ugh you're right. It's loading twice. I'll have to do it inline.

chrisarcand:
@miq-bot add_label wip

chrisarcand:
Yeah this just isn't going to work without some extra elbow grease. I've updated the issue with my findings.

PR #316, opened 2017-02-15
Add path based labeler

chrisarcand:
~~**REQUIRES https://github.com/ManageIQ/miq_bot/pull/321** Merge that first for a proper diff.~~

Adds a generic labeler that labels things by matching the names of changed files in pull requests to whatever regular expression you want.

* We can now automatically add the Ansible Tower label for obvious paths to Ansible stuff (app/models/manageiq/providers/ansible_tower...). This is prolly temporary until it's extracted to it's own gem, but... people keep forgetting labels during our crunch time, so here you go!
* Replacement for #312 (label changes to graphic files)
* Replaces `SqlMigrationLabeler`
* Replaces `GemChangesLabeler`

One thing to note is that this will cause a sort of annoying Settings.yml as you might need to specify a common pattern (like sql migrations, etc) as a separate rule for every repo you want it on. That's dumb, but this is big enough as it is so let's save optimizing settings for another time.


chrisarcand:
🎁  @chessbyte 

@Fryguy Thoughts?

app/workers/concerns/branch_worker_mixin.rb
```diff
@@ -51,9 +51,7 @@ def diff_details_for_merge
     end
   end
 
-  def diff_file_names_for_merge
-    branch.repo.with_git_service do |git|
-      git.diff_file_names(branch.local_merge_target, commits.last)
-    end
+  def diff_file_names
+    branch.git_service.diff.new_files
```
chrisarcand: `diff_file_names_for_merge` is unused and looks to be old, old Minigit stuff. I repurposed it for my uses here and tried to use the current Rugged interface. Looks...sorta bad but I know the interface is a work in progress. So lemme know if there's a cleaner way to get file names in the diff ¯\\\_(ツ)_/¯

Fryguy: This is good :+1:

Fryguy:
Cool idea!

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
```diff
@@ -0,0 +1,27 @@
...
+    return unless verify_branch_enabled
+
+    process_branch
+  end
+
+  private
+
+  def process_branch
+    label_rules.each do |rule|
+      if diff_file_names.any? { |file_name| file_name =~ Regexp.new(rule.pattern) }
```
Fryguy: You might want to memoize the diff_file_names, so it's not re-executing the rugged command over and over.

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
```diff
@@ -0,0 +1,27 @@
...
+        GithubService.add_labels_to_an_issue(fq_repo_name, pr_number, [rule.label])
```
Fryguy: Prefer collecting all of the labels up front, then telling the GithubService to add all of the labels at once.

Also, previously, we had a method that would add only the labels that were a) new and b) existing.  This was because the GitHub API would create a label even if it didn't exist, which is not something we wanted it to do.  I'm not sure where that method went...I assume it was removed as part of the last refactoring?

config/settings.yml
```diff
@@ -27,6 +27,11 @@ github_notification_monitor:
   repo_names: []
 merge_target_titler:
   enabled_repos: []
+path_based_labeler:
+  enabled_repos: []
+  rules: []
+    # - pattern:
+    #   label:
```
Fryguy: We might want different rules per repo.  This is because the labels for one repo are not likely to be the same for other repos.  This is, in fact, already the case with ManageIQ/manageiq versus ManageIQ/manageiq-ui-classic, for example.

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
chrisarcand: Good call.

Just moved: https://github.com/ManageIQ/miq_bot/blob/637e49616839c5765d9c2f050c43338afc396f04/lib/github_service.rb#L52-L62

config/settings.yml
chrisarcand: I was just about to comment on this given the other PR ^_^

So I started that way but ran in to the `enabled_repos` behavior, wherever that is coming from. I wasn't sure how to structure the yml in a way that made sense. What do you expect the yml to look like with these rules and `enabled_repos`?

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
Fryguy: @chrisarcand I just updated my message above to be clearer...Can you read again?

config/settings.yml
Fryguy: enabled_repos is the old enable/disable functionality that I'm trying to revamp and make more common.  For now it has to stay though.

Even though it's kind of clunky, perhaps

```yaml
path_based_labeler:
  enabled_repos:
    ManageIQ/manageiq:
    - pattern:
       label:
    - pattern:
       label:

File: app/workers/pull_request_monitor_handlers/path_based_labeler.rb (hunk 3 above)
Comment (chrisarcand) [inline app/workers/pull_request_monitor_handlers/path_based_labeler.rb]: I believe that logic is just moved in to GithubService::Issue now:

https://github.com/manageiq/miq_bot/blob/637e49616839c5765d9c2f050c43338afc396f04/lib/github_notification_monitor.rb#L132

and

https://github.com/manageiq/miq_bot/blob/637e49616839c5765d9c2f050c43338afc396f04/lib/github_service/issue.rb#L21-L23

The difference for our case here is, unfortunately, that we're not dealing with the already-fetched issue response from Github but processing a branch. 

Will have to ponder further 😕

File: config/settings.yml (hunk 4 above)
Comment (chrisarcand) [inline config/settings.yml]: I think I'll still have to tweak [the check](https://github.com/ManageIQ/miq_bot/blob/637e49616839c5765d9c2f050c43338afc396f04/app/models/repo.rb#L72-L74) to account for that different format but that shouldn't be terribly difficult. Alrighty.

Comment (Fryguy) [inline config/settings.yml]: Then we just have to update Repo#enabled_for? to:

```ruby
  def enabled_for?(checker)
    enabled_repos = Settings[checker].enabled_repos || []
    enabled_repos = Array(enabled_repos) if enabled_repos.kind_of?(String)
    enabled_repos.include?(name)
  end
```

There might be a more elegant way to do this, but it's eluding me.

These tests should all pass (which I think all of the above does)

- enabled_repos: nil
- enabled_repos: single_string
- enabled_repos: ["array", "of, "string"]
- enabled_repos: {"hash" => "of, "key" => "values"}

chrisarcand: ^_^

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
Fryguy: Oh I see...in that case we already had the fetched labels, but here we would have to issue a query to find all of the existing labels.

Maybe we should re-verify the old assumptions above.  Perhaps the GitHub API was corrected to no longer do this.

These two tests should verify:

- `GithubService.add_labels_to_an_issue(fq_repo_name, pr_number, ["foo"])` and see if it creates a new label in the repo or raises an issue.  If it still does the former, we'll have to deal with that potentially (or maybe it's not an issue since this configuration is not user-editable)
- Create an issue, add the bug label, then call: `GithubService.add_labels_to_an_issue(fq_repo_name, pr_number, ["bug"])` and see if it makes a notification in the UI about adding a label even if it already exists.  If it still does that, we will most definitely have to deal with it otherwise everytime someone does a git push, you will get that "added a label" notification

Fryguy: Thinking about this...since GithubService::Issue is a reusable class now, perhaps we just fetch that issue, then call add_labels directly on that?

chrisarcand: > Thinking about this...since GithubService::Issue is a reusable class now, perhaps we just fetch that issue, then call add_labels directly on that?

That entirely depends on your previous comment; I didn't think so at first given that you'd then be doing a fetch before doing something that I *assumed* was idempotent (but, as you said, I should probably verify first 😅)

Fryguy: Right, yeah, I still think we should double check the old assumptions regardless.

chrisarcand: * Ugh. `GithubService.add_labels_to_an_issue` *does* create a new label if it doesn't exist. As you mentioned, for the GHNM it's fine because we check first (because we already have the issue!) but here would require another call. I'm tempted to say whatever, because as you said this isn't user configurable at all. We should still add our own method to GithubService as I'm sure we'll need the 'add label only if exists' functionality eventually anyway (not this PR).

* Adding a label to an issue that already had that label *is* idempotent, thankfully. No notification, no UI update of any variety.

chrisarcand: Yeah I'm not going to care about this here. It's our own fault if we configure the bot with a misspelled tag, not worth the verification call.

app/workers/concerns/branch_worker_mixin.rb
```diff
@@ -51,9 +51,7 @@ def diff_details_for_merge
     end
   end
 
-  def diff_file_names_for_merge
-    branch.repo.with_git_service do |git|
-      git.diff_file_names(branch.local_merge_target, commits.last)
-    end
+  def diff_file_names
+    @diff_file_names ||= branch.git_service.diff.new_files
```
chrisarcand: I *think* this memoization is ok.

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
chrisarcand: https://github.com/ManageIQ/miq_bot/pull/316/files#diff-9bb1d9f485db689f63f34a7fda93116fR54

chrisarcand:
@Fryguy This is ready for an other glance.

chrisarcand:
@Fryguy Wait, my bad, still have to handle "Prefer collecting all of the labels up front, then telling the GithubService to add all of the labels at once."

app/models/repo.rb
```diff
@@ -70,7 +70,9 @@ def with_travis_service
   end
 
   def enabled_for?(checker)
-    Array(Settings.public_send(checker).enabled_repos).include?(name)
+    Array(Settings.public_send(checker).enabled_repos).each_with_object([]) { |value, values|
+      values << (value.kind_of?(Array) ? value.first.to_s : value)
+    }.include?(name)
```
Fryguy: `.each_with_object([])` is an immediate code smell to me:

```ruby
    Array(Settings.public_send(checker).enabled_repos).collect do |value|
      value.kind_of?(Array) ? value.first.to_s : value
    end.include?(name)
```

app/workers/pull_request_monitor_handlers/path_based_labeler.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+    end
+  end
+
+  def label_rules
+    Settings.path_based_labeler.enabled_repos.public_send(fq_repo_name)
```
Fryguy: Prefer square bracket access over public_send for dynamic stuff (I'm not sure why enabled_for? was ever written that way, but it's not the preferred way)

```ruby
Settings.path_based_labeler.enabled_repos[fq_repo_name]
```

Fryguy:
@chrisarcand Does this also kill some of the other things like sql_migration_labeler and gemfile_labeler?

app/models/repo.rb
chrisarcand: You can't do that as not all objects returned from `enabled_repos` are Enumerable. That was my first implementation :)

Code smell? `each_with_object` is just `reduce` yielding the memo value at the end of the block automatically. So you'd be saying that the higher order function `reduce` is a code smell... 😕

chrisarcand:
>  Does this also kill some of the other things like sql_migration_labeler and gemfile_labeler?

Yes I believe it does, left those out of this one but let's for sure kill them.

app/models/repo.rb
Fryguy: No, the code smell is that 9/10 times when people use each_with_object with an Array, it can just be changed into a regular method call on Array, like .collect.

chrisarcand: Oh hah, I could see that. In my [lesser] experience I just don't see people thinking to use reduction and never see anyone using it. Fair.

Anyway, leaving this here given my point above since this seems a valid use case, otherwise if you still insist we can just use moar intermediate lines.

chrisarcand:
I just realized that this should be a CommitMonitor handler, not a pull request one, especially given that most of PullRequestMonitor will be torn out with webhooks.

chrisarcand:
@miq-bot add_label wip

chrisarcand:
@miq-bot remove_label wip

chrisarcand:
@Fryguy This is ready. It also has a lot of changes since you last looked, so see updated description and added code.

app/workers/commit_monitor_handlers/branch/path_based_labeler.rb
```diff
@@ -0,0 +1,34 @@
...
+  private
+
+  def process_branch
+    labels = []
+    label_rules.each do |rule|
+      pattern = Regexp.new(rule.pattern)
+      if diff_file_names.any? { |file_name| file_name =~ pattern }
+        labels << rule.label
+      end
+    end
```
Fryguy: Minor refactor idea (no need to actually change for this PR)

```ruby
  def process_branch
    labels = label_rules.collect do |rule|
      pattern = Regexp.new(rule.pattern)
      diff_file_names.any? { |file_name| file_name =~ pattern }
    end.collect(&:label)

    GithubService.add_labels_to_an_issue(fq_repo_name, pr_number, labels) if labels.present?
  end
```

PR #317, opened 2017-02-15
Replace blob getter with a content getter

Fryguy:
@chrisarcand or @bdunne Please review.

This is a minor refactor, which I need for enable_disable support, but could be pulled out into its own PR.  Basically, I can't see why a caller would ever want the raw Rugged::Blob object versus just its content.  So, this changes the interface to just get the content.  The only current caller also just wants the content.

### DIFF
--- a/lib/git_service/branch.rb
+++ b/lib/git_service/branch.rb
@@ -7,12 +7,8 @@ module GitService
       @branch = branch
     end
 
-    def blob_at(path) # Rugged::Blob object for a given file path on this branch
-      blob_data = tip_tree.path(path)
-      blob = Rugged::Blob.lookup(rugged_repo, blob_data[:oid])
-      (blob.type == :blob) ? blob : nil
-    rescue Rugged::TreeError
-      nil
+    def content_at(path)
+      blob_at(path).try(:content)
     end
 
     def diff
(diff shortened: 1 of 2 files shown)


Fryguy:
Wait on merge...I missed a file during commit.

Fryguy:
Ok, this is ready to go now

lib/git_service/branch.rb
```diff
@@ -7,12 +7,8 @@ def initialize(branch)
...
     end
 
-    def blob_at(path) # Rugged::Blob object for a given file path on this branch
-      blob_data = tip_tree.path(path)
-      blob = Rugged::Blob.lookup(rugged_repo, blob_data[:oid])
-      (blob.type == :blob) ? blob : nil
-    rescue Rugged::TreeError
-      nil
+    def content_at(path)
+      blob_at(path).try(:content).try(:to_s)
```
chrisarcand: Is the `to_s` necessary? Looks like it should [already come back as a String](http://www.rubydoc.info/github/libgit2/rugged/master/Rugged/Blob#content-instance_method)

chrisarcand:
The only other thing that might be useful from a publicly available blob is maybe `binary?` but...whatever until that's actually needed. Looks good,  just one comment.

lib/git_service/branch.rb
Fryguy: Good call...I only added it because the only caller we had it, but yeah doesn't look necessary.

Fryguy:
@chrisarcand Updated

PR #318, opened 2017-02-15
Standardize on enable/disable of repos across all workers

Fryguy:
This PRs standardizes on repo names in the settings.yml for all workers and supports them in any worker.

The way the settings.yml will work is to specify either included_repos, excluded_repos, or nothing.

```yaml
worker1:  # Will run in all repos

worker2:  # Will only run in the specified repos
  included_repos:
  - "org1/repo1"
  - "org2/repo12"

worker3:  # Will run in all repos except the specified repos
  excluded_repos:
  - "org1/repo1"
  - "org2/repo12"

worker4:  # Will raise an exception, since you should not specify both
  included_repos:
  - "org1/repo1"
  excluded_repos:
  - "org2/repo12"

worker5:  # Effectively disables the worker
  included_repos: []
```

TODO:

- [x] Specs.
- [x] Determine if find_branch should also check for enablement.  It's not necessary because the workers will not be queued, but could potentially be queued outside of the workflow.
- [x] Update the default settings.yml for each worker with sensible defaults (i.e. default to all repos, or default to no repos)
- Support enable/disable for specific linters. (deferring to a separate PR)
- Support yaml references to avoid a lot of copy/paste in the yaml.  (deferring to a separate PR)
    Example:

    ```yaml
    core_repos: &core_repos
    - ManageIQ/manageiq
    - ManageIQ/manageiq-ui-classic
    - ManageIQ/manageiq-ui-service
    - ManageIQ/manageiq-providers-amazon
    ... # 10 more repos

    worker1:
      included_repos:
      - *core_repos
    worker2:
      included_repos:
      - *core_repos
      - ManageIQ/additional_repo
    ```

### DIFF
--- a/app/workers/concerns/sidekiq_worker_mixin.rb
+++ b/app/workers/concerns/sidekiq_worker_mixin.rb
@@ -5,10 +5,64 @@ module SidekiqWorkerMixin
   extend ActiveSupport::Concern
 
   included do
-    delegate :sidekiq_queue, :workers, :running?, :to => self
+    delegate :settings, :enabled_repos, :enabled_repo_names, :enabled_for?, :to => :class
+    delegate :sidekiq_queue, :workers, :running?, :to => :class
   end
 
   module ClassMethods
+    #
+    # Settings helper methods
+    #
+
+    def settings_key
+      @settings_key ||= name.split("::").last.underscore
+    end
+    private :settings_key
+
+    def settings
+      Settings[settings_key] || Config::Options.new
+    end
+
+    def included_and_excluded_repos
+      i = settings.included_repos.try(:flatten)
+      e = settings.excluded_repos.try(:flatten)
+      raise "Do not specify both excluded_repos and included_repos in settings for #{settings_key.inspect}" if i && e
+      return i, e
+    end
+    private :included_and_excluded_repos
+
+    def enabled_repos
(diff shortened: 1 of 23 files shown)


app/workers/concerns/sidekiq_worker_mixin.rb
```diff
@@ -5,10 +5,64 @@ module SidekiqWorkerMixin
...
+    private :included_and_excluded_repos
+
+    def enabled_repos
+      i, e = included_and_excluded_repos
+
+      if i && !e
+        Repo.where(:name => i)
+      elsif !i && e
+        Repo.where.not(:name => e)
+      elsif !i && !e
```
bdunne: Shouldn't this just be `else`?

app/workers/concerns/sidekiq_worker_mixin.rb
```diff
@@ -5,10 +5,64 @@ module SidekiqWorkerMixin
...
+    end
+
+    def enabled_for?(repo)
+      i, e = included_and_excluded_repos
+
+      if i && !e
+        i.include?(repo.name)
+      elsif !i && e
+        !e.include?(repo.name)
+      elsif !i && !e
```
bdunne: same here

PR #321, opened 2017-02-16
Safer labeling

chrisarcand:
This does a few things:

* `remove_labels` is no longer supported, because the API does not support
  multiple label removal. The trick around that before was to replace all
  the labels on the issue instead of removing one by one, but that causes
  a new 'label adding event' in the UI even as it's replacing. Lame.
  Forunately we weren't using the plural form anywhere yet anyway.

* Consolidates all of the logic of "Don't add a label to an issue if
  that issue already has the label" all in one spot, and funnels ALL
  calls (through the Issue wrapper OR the GithubService call) through
  the same interface. The Issue class has the logic because if you put
  it in the overrided method in GithubService, you'd paying a tax in
  refetching the issue if the Issue object is trying to add a label to
  itself.

* Adds clarification everywhere that the two add_label interfaces expect
  callers to validate the presence of labels on a repository. The reason
  for this is best explained by example: The Notification Monitor
  responds to users what labels weren't valid. We don't need to
  revalidate labels internally past that point. So we assume that if
  you're using that API you know what you're doing. Also saves calls for
  things that a guaranteed to be 'correct' (non-user inputted labels)

  The benefits include not pulling our hair out when
  the API does funny things depending on where we call adding labels,
  code consolidation, etc. The biggest benefit is that when labels are
  attempted to be added from the bot that already exist on the issue, it
  isn't reported as an event (Github shows labels being added in the UI
  even when they are already added. wtf?)

  The drawbacks include more API calls on every
  single label action - but given the amount of time figuring out
  the little quirks of every way to do this and that we are well below
  our rate limit, AND that we expect to be even further away from the
  rate limit with webhooks, I think this safety is better. Also remember
  these sorts of things scale with user interaction, not number of repos
  or any junk like that.

### DIFF
--- a/lib/github_service.rb
+++ b/lib/github_service.rb
@@ -50,31 +50,22 @@ module GithubService
     end
 
     # Overrides Octokit.add_labels_to_an_issue
-    # Github automatically creates a label for the repo if it does not exist.
-    # This requires you to opt-in to that behavior.
-    def add_labels_to_an_issue(fq_repo_name, issue_number, requested_labels, create_if_absent: false)
-      requested_labels = Array(requested_labels)
-      labels_to_add = requested_labels
-
-      unless create_if_absent
-        requested_labels.each do |label|
-          unless valid_label?(fq_repo_name, label)
-            Rails.logger.warn "WARNING: The label '#{label}' was attempted to be added to #{fq_repo_name} ##{issue_number}, but that label doesn't exist on the repo."
-            labels_to_add.delete(label)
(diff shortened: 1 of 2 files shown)


chrisarcand:
Final labeling PR (I hope). Sorry for all the words but I think it's important to document every last thing so we can reference later.

This should fix some of the funny business with the bot re-adding labels that are already added, and fixes future work for the path based labeler.

PR #325, opened 2017-02-17
Actual regex in path based labeler settings

chrisarcand:


chrisarcand:
@Fryguy

chrisarcand:
😏  https://github.com/ManageIQ/guides/pull/191

PR #328, opened 2017-02-28
Refactor Github command structure

chrisarcand:
This makes it pretty easy to plug in web hooks instead of notifications. Removes the command logic from GHNM and builds a command registry to dispatch commands to.

This is a large PR, I know, but if you view commit by commit it's much easier to digest. Doing half and half registry + how it's currently done is actually harder to consume.


chrisarcand:
Addressed bot comments, except for UnusedMethodArgument as I want those explicitly stated in the interface signature because that's how every subclass must implement it.

PR #311, opened 2017-02-14
Issue mover

chrisarcand:
**REQUIRES #329; merge that first**

Closes #258

@Fryguy Please review

cc/ @blomquisg 

Move success (original issue):

Move success (new issue):

Trying to move a pull request:

Trying to move an issue and you're not part of the org:

With several qualifications, errors are combined to human readable sentences so people don't hit their head against the wall over and over. For example, trying to move a pull request AND trying to specify a repo outside of the issue's organization:

PR #329, opened 2017-02-28
Add command restrictions

chrisarcand:
This makes it possible to easy restrict a command to a subset of users. The only type of restriction currently supported is organization, which restricts use of those commands to members of the issue's organization.

**REQUIRES #328, merge that first for a sane diff**

Enables #311, once the issue mover is properly rebased with the new command structure.

### DIFF
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -1,7 +1,9 @@
 module GithubService
   module Commands
     class AddLabel < Base
-      def execute!(issuer:, value:)
+      private
+
+      def _execute(issuer:, value:)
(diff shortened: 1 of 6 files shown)


Fryguy:
As discussed, the `super` feels weird, and it might just be cleaner to rename the `execute!` method in the subclasses to `raw_execute`.  If you're concerned a caller might use it directly, you can just make it private or just put some documentation on it saying that the preferred calling method is `execute!`

chrisarcand:
@Fryguy How's that?

chrisarcand:
GAAAAH NOT HASH ROCKETS AGAAAAAAAIN

PR #336, opened 2017-03-29
Rescue UnmergeableError

bdunne:
While trying to generate the diff on an unmergeable PR it will raise.
If this is not rescued, the job will be requeued and continue to fail

image


bdunne:
@Fryguy Please review

PR #346, opened 2017-05-25
Add listener for configuration changes

Fryguy:
Use the listen gem to listen for filesystem changes to config files.
This will allow us to modify configuration on the fly, and have those
updates reflected in the Rails and Sidekiq processes.

@bdunne Please review
cc @chessbyte


Fryguy:
I'm not sure how to solve this problem since we commit Gemfile.lock... @bdunne Thoughts?

chrisarcand:
Is this really worth it to save one extra command to restart the bot each time?

[I say no, but...] Looks like if you want to fix this, it's a known issue on Travis, Gemfile.lock shouldn't be a problem: https://github.com/guard/listen#important-if-you-cannot-install-listen-eg-on-travisci-builds-a-workaround-is-here-httpsgithubcomguardlistenwikiruby-version-requirements

Fryguy:
> Is this really worth it to save one extra command to restart the bot each time?

It's not the difficulty of rebooting the bot that's the problem, it's that we have to reboot the bot at all for a config change.  This requires us to manually watch the log file so we can pick an opportune time to shut the bot down (e.g. when it's not in the middle of a git checkout), and is prone to error.  There's no need to bounce the bot for this common case.

chrisarcand:
> This requires us to manually watch the log file so we can pick an opportune time to shut the bot down

Yeah that's fair, good point.

Fryguy:
@bdunne Thoughts? https://github.com/ManageIQ/miq_bot/pull/346#issuecomment-304075783

bdunne:
Bundle on rhel / centos?

chrisarcand:
I think I'm missing something. What's the problem here?

bdunne:
@chrisarcand Some dependencies are different on mac vs linux.  See https://travis-ci.org/ManageIQ/miq_bot/builds/236112238#L434

chrisarcand:
@bdunne Are you sure? I think this is just that Travis is using 2.2.0 when "2.2" is specified in the .travis-yml. ruby_dep [frustratingly] constantly expects you to use the most up-to-date Ruby version and fails to install if not.

It looks like if you instruct Travis to use 2.2.6 all is well:

https://github.com/chrisarcand/miq_bot/commit/0c755d96cec21228fe819c28d3eb5cd486729c3b
https://travis-ci.org/chrisarcand/miq_bot/builds/239664436

We use 2.2.6 on the live bot anyway.

Orrrr we could have just had a `.ruby-version` but we've been down this path 4 times before ;)

Fryguy:
Bumped with the .travis.yml change...let's see how it goes.

chrisarcand:
🎉

PR #358, opened 2017-09-26
Fix irreverible migration

imtayadeway:
Trying to revert this migration results in this error:

```
== 20170127173128 FixLimitInSchema: reverting =================================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

ActiveRecord::IrreversibleMigration/home/tim/src/miq_bot/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7/lib/active_record/migration/comman
d_recorder.rb:65:in `inverse_of'
```

~~Changing it to an `up` and doing nothing on the `down` should be fine.~~

Edit: We need to tell it explicitly how to undo the change by setting
`:limit => nil`.

@miq-bot assign @chrisarcand

### DIFF
--- a/db/migrate/20170127173128_fix_limit_in_schema.rb
+++ b/db/migrate/20170127173128_fix_limit_in_schema.rb
@@ -1,5 +1,5 @@
 class FixLimitInSchema < ActiveRecord::Migration
-  def change
+  def up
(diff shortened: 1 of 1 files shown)


imtayadeway:
@chrisarcand you are correct 😁 Just updated!

PR #359, opened 2017-09-28
Add a YAML linter

imtayadeway:
Warning: requires a `yamllint` executable on the path of the bot. It's available in dnf/apt/etc... but I'd recommend using the most recent version available via `pip`.

See https://github.com/miq-test/sandbox/pull/42#issuecomment-332899807 for an example of output.

@miq-bot assign @Fryguy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 cache: bundler
 services:
 - redis-server
+before_install:
+- sudo pip install yamllint
 before_script:
 - cp config/database.tmpl.yml config/database.yml
 - bundle exec rake db:setup
--- a/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
@@ -42,7 +42,7 @@ class CommitMonitorHandlers::CommitRange::RubocopChecker::MessageBuilder
   end
 
   def header
-    header1 = "Checked #{"commit".pluralize(commits.length)} #{commit_range_text} with ruby #{RUBY_VERSION}, rubocop #{rubocop_version}, and haml-lint #{hamllint_version}"
+    header1 = "Checked #{"commit".pluralize(commits.length)} #{commit_range_text} with ruby #{RUBY_VERSION}, rubocop #{rubocop_version}, haml-lint #{hamllint_version}, and yamllint #{yamllint_version}"
 
     file_count    = results.fetch_path("summary", "target_file_count").to_i
     offense_count = results.fetch_path("summary", "offense_count").to_i
@@ -123,5 +123,10 @@ class CommitMonitorHandlers::CommitRange::RubocopChecker::MessageBuilder
 
   def hamllint_version
     HamlLint::VERSION
+  end
+
+  def yamllint_version
+    _out, err, _ps = Open3.capture3("yamllint -v")
+    err.split.last
   end
 end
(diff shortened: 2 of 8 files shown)


Fryguy:
Any way to make these not crash boom bang?  maybe ⚠️ ?

app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
```diff
@@ -124,4 +124,9 @@ def rubocop_version
   def hamllint_version
     HamlLint::VERSION
   end
+
+  def yamllint_version
+    _out, _err, status, _thread = Open3.popen3("yamllint -v")
```
Fryguy: I think the variables are named incorrectly, and also it doesn't closed the in and out pipes, so I don't think this is what you want to do...

According to the [docs](https://ruby-doc.org/stdlib-2.4.2/libdoc/open3/rdoc/Open3.html#method-c-popen3), popen3 returns `stdin, stdout, stderr, wait_thr`.  So I gather you are just trying to read the version from STDERR.  If so, it's probably preferable to use capture3:

```ruby
_out, err, _ps = Open3.capture3("yamllint -v")
err.chump.split.last
```

imtayadeway: This is what happens when I scan three different articles on how to do this 😆

imtayadeway:
> Any way to make these not crash boom bang? maybe ⚠️ ?

@Fryguy I think the best way to do this would be either to set the level for each rule (I've set the indentation rule level to warning in my sandbox PR to see how this looks), or to simply extend from the relaxed set of rules (it's currently on default), which should have the same effect.

imtayadeway:
@Fryguy WDYT about [this report](https://github.com/miq-test/sandbox/pull/42#issuecomment-332993199) based on [this configuration](https://github.com/miq-test/sandbox/pull/42/files#diff-b97c55dd898003930e2858fadd3b5c9c)? The crash boom bang seems appropriate for a duplicate key violation - I'd expect most offenses to get a ⚠️ as with the indentation offense. Now to figure out those errors on travis....

Fryguy:
Oh that's great!  Yeah, we may have to be more prescriptive per rule, but we can always change that as we go.

---

Incidentally, this string is getting unwieldy:

    ruby 2.3.1, rubocop 0.47.1, haml-lint 0.20.0, and yamllint 1.8.1

We may want to drop that into a `<details></details>` tag instead to let GitHub collapse it.  Thoughts?  Obviously not for this PR, but I can open a separate issue.

Example:

```
<details>
  <summary>Versions</summary>

- ruby 2.3.1
- rubocop 0.47.1
- haml-lint 0.20.0
- yamllint 1.8.1

</details>
```

produces:

<details>
  <summary>Versions</summary>

- ruby 2.3.1
- rubocop 0.47.1
- haml-lint 0.20.0
- yamllint 1.8.1

</details>

Fryguy:
Looks like you need to install yamllint in travis.  I *think* you just have to do `pip install yamllint` in the .travis.yml in the before_install section.

.travis.yml
```diff
@@ -4,6 +4,8 @@ sudo: false
 cache: bundler
 services:
 - redis-server
+before_install:
+- sudo pip install yamllint
```
bdunne: Is this the same as the `yamllint` gem?

imtayadeway: From the description of the yamllint gem:

```
Checks YAML files for correct syntax. Currently it checks for:

    Valid YAML syntax
    Overlapping key definitions in YAML files, where the last definition would win
```

It would be much better if we could use a rubygem, and I had a lot of hesitancy introducing a python package into the bot! But I haven't seen a better solution - it does a lot more than the gem as it also checks indentation, trailing whitespace, whitespace consistency etc...

PR #365, opened 2017-10-06
Implement a daily Code analyzer

bdunne:
I hope this will be added to the statistics on the Sprint review. Here we can have an overall metric on how many linter offenses there are in each repo.  Eventually I'd like to keep historical data, but on this first pass, we can get daily numbers saved on the branch record.

### DIFF
--- a/app/workers/commit_monitor_handlers/branch/code_analyzer.rb
+++ b/app/workers/commit_monitor_handlers/branch/code_analyzer.rb
@@ -21,17 +21,8 @@ class CommitMonitorHandlers::Branch::CodeAnalyzer
 
   def analyze
     branch.repo.git_fetch
-    run_linters
+    @results = merge_linter_results(run_all_linters)
     offense_count = @results.fetch_path("summary", "offense_count")
     branch.update_attributes(:linter_offense_count => offense_count)
-  end
-
-  def run_linters
-    unmerged_results = run_all_linters
-    if unmerged_results.empty?
-      @results = {"files" => []}
-    else
-      @results = merge_linter_results(*unmerged_results)
-    end
   end
 end
--- a/app/workers/commit_monitor_handlers/branch/code_analyzer.rb
+++ b/app/workers/commit_monitor_handlers/branch/code_analyzer.rb
@@ -1,0 +1,37 @@
+require 'rugged'
+
+class CommitMonitorHandlers::Branch::CodeAnalyzer
+  include Sidekiq::Worker
+  sidekiq_options :queue => :miq_bot_glacial
+
+  include BranchWorkerMixin
+  include ::CodeAnalysisMixin
+
+  def self.handled_branch_modes
+    [:regular]
+  end
+
+  def perform(branch_id)
+    return unless find_branch(branch_id, :regular)
+
+    analyze
+  end
+
+  private
+
+  def analyze
+    branch.repo.git_fetch
+    run_linters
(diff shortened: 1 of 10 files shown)


app/workers/commit_monitor_handlers/branch/code_analyzer.rb
```diff
@@ -0,0 +1,37 @@
...
+    branch.update_attributes(:linter_offense_count => offense_count)
+  end
+
+  def run_linters
+    unmerged_results = run_all_linters
+    if unmerged_results.empty?
+      @results = {"files" => []}
+    else
+      @results = merge_linter_results(*unmerged_results)
+    end
```
Fryguy: This conditional feels like it could live in CodeAnalysisMixin

app/workers/schedulers/code_analyzer.rb
```diff
@@ -0,0 +1,25 @@
...
+    sidekiq_options :queue => :miq_bot_glacial, :retry => false
+
+    include Sidetiq::Schedulable
+    recurrence { daily }
+
+    include SidekiqWorkerMixin
+
+    def perform
+      Repo.where(:name => fq_repo_names).each do |repo|
+        repo.branches.pluck(:id).each do |branch_id|
```
Fryguy: I think we only want to run this on non-PR branches.

lib/git_service/branch.rb
```diff
@@ -53,8 +53,24 @@ def target_for_reference(reference) # Rugged::Commit for a given refname i.e. "r
...
+      recursive_list_files_in_tree(tip_tree.oid)
+    end
+
     private
 
+    def recursive_list_files_in_tree(rugged_tree_oid, files = [], current_path = Pathname.new(""))
+      rugged_repo.lookup(rugged_tree_oid).each do |i|
+        case i[:type]
+        when :blob
+          files << current_path.join(i[:name]).to_s
```
Fryguy: You can extract `current_path.join(i[:name])` to a variable since it's in both paths

lib/git_service/branch.rb
```diff
@@ -53,8 +53,24 @@ def target_for_reference(reference) # Rugged::Commit for a given refname i.e. "r
       rugged_repo.references[reference].target
     end
 
+    def tip_files
+      recursive_list_files_in_tree(tip_tree.oid)
+    end
+
     private
 
+    def recursive_list_files_in_tree(rugged_tree_oid, files = [], current_path = Pathname.new(""))
```
Fryguy: I think there's way to write this where the files doesn't have to be passed all the way into the recursion.  That is, make the method more functional-like in that it returns the Array for a given path, and then it's just concating Arrays up the stack...something like:

```ruby
def recursive_list_files_in_tree(rugged_tree_oid, current_path = Pathname.new(""))
  rugged_repo.lookup(rugged_tree_oid).each_with_object([]) do |i, files|
    next_path = current_path.join(i[:name])
    case i[:type]
    when :blob
      files << next_path.to_s
    when :tree
      files.concat(recursive_list_files_in_tree(i[:oid], next_path))
    end
  end
end
```

Fryguy: Upon re-reading what I wrote, you can probably remove the `recursive_` bit, since that's an implementation detail.

Fryguy:
Really nice idea!  Do you know how long it takes to run?

bdunne:
Okay @Fryguy I think I addressed all of your comments

PR #366, opened 2017-10-10
Avoid having :pull_request => nil

bdunne:


chrisarcand:
Ugh, I'm sad we use this gem here, too 😩

bdunne:
For this case, you could change rails :)

Fryguy:
Yeah, I don't see why we can't use column defaults here for most stuff (except maybe the serialized arrays)

Fryguy:
> Ugh, I'm sad we use this gem here, too 😩

This was a playground for the usage of that gem, so technically this was first :)

PR #370, opened 2017-11-13
New comment issues

bdunne:
- Don't blow up when a comment body is nil
  - Example: https://github.com/ManageIQ/manageiq-appliance/pull/147
- Enhance `issue_number` matching to verify that everything after `/` is an integer
- Only process notifications that have an `issue_number`


bdunne:
Maybe the Github API changed?

PR #371, opened 2017-11-13
Sync the Gemfile.lock from the running system

bdunne:


chrisarcand:
How did that happen in the first place? 😕

bdunne:
I believe we did a `bundle update` when we updated the ruby on the system.

PR #372, opened 2017-11-13
Linting fixes

bdunne:


app/workers/concerns/code_analysis_mixin.rb
```diff
@@ -22,6 +23,6 @@ def run_all_linters
     unmerged_results << Linter::Rubocop.new(branch).run
     unmerged_results << Linter::Haml.new(branch).run
     unmerged_results << Linter::Yaml.new(branch).run
-    unmerged_results.compact!
+    unmerged_results.compact
```
NickLaMuro: Instead of creating a new array, can't you just one of the following:

```ruby
# Option 1
unmerged_results.compact!
unmerged_results

# Option 2
unmerged_results.tap(&:compact!)
```

PR #373, opened 2017-11-20
Address CVE-2017-8418 by upgrading Rubocop gem

chessbyte:
Requested in https://github.com/ManageIQ/manageiq/pull/16491


chrisarcand:
@chessbyte It looks like you ran `bundle update` which updates all the gems to the latest possible within the bundle; I'd probably just keep it to the scope of the PR by removing that last commit and just running `bundle update rubocop` instead

chrisarcand:
It looks like in the new rubocop version they added the RubyDoc link to found offenses, and we have a test that expects the old message format in testing the bot. The test will have to be updated.

Fryguy:
> It looks like in the new rubocop version they added the RubyDoc link to found offenses

That's awesome!  Maybe I can remove that code I have to generate the link.

PR #374, opened 2017-11-27
Updated dependencies for ruby 2.4 compatibility

europ:


Fryguy:
@bdunne Wasn't there a reason we didn't go to Ruby 2.4 on the production bot?

bdunne:
Yeah, I think we wanted to stay aligned with the version of ruby we're running on ManageIQ.  Also, I think there were other problems with ruby 2.4 when we discussed it last.

Fryguy:
> Yeah, I think we wanted to stay aligned with the version of ruby we're running on ManageIQ.

Oh right...we need the bot to run at the minimum for ManageIQ which is 2.3.1

skateman:
@Fryguy @bdunne there were some openssl compilation issues with ruby 2.2 on both @europ's deb-based distro and my Fedora.

Fryguy:
we use ruby 2.3 on the bot anyway, so not sure why 2.2 was being installed.

skateman:
@europ change your local setup to 2.3 and update the PR accordingly. I'm sure the guys above will like it more :wink:

Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+gem 'xmlrpc'
```
Fryguy: For future reference @chrisarcand @bdunne ... This is required for Ruby 2.4 because it was ripped out of stdlib and into a gem.

Fryguy: And it actually belongs in active_bugzilla.

chrisarcand: yup!

europ:
I have an issue with ruby 2.3.1 `undefined symbol: SSLv2_method` after executing command `rbenv install 2.3.1`.

```
Downloading ruby-2.3.1.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.1.tar.bz2
Installing ruby-2.3.1...

BUILD FAILED (Ubuntu 16.04 using ruby-build 20170523-25-g476d09b)

Inspect or clean up the working tree at /tmp/ruby-build.20171129215048.12124
Results logged to /tmp/ruby-build.20171129215048.12124.log

Last 10 log lines:
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/package.rb:44:in `<top (required)>'
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/installer.rb:10:in `<top (required)>'
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /tmp/ruby-build.20171129215048.12124/ruby-2.3.1/lib/rubygems/core_ext/kernel_require.rb:55:in `require'
	from ./tool/rbinstall.rb:657:in `<module:RbInstall>'
	from ./tool/rbinstall.rb:585:in `<main>'
uncommon.mk:260: recipe for target 'do-install-all' failed
make: *** [do-install-all] Error 1
```

[ruby-build.20171129215048.12124.log](https://github.com/ManageIQ/miq_bot/files/1515514/ruby-build.20171129215048.12124.log)

I tried to fix it, but without success. There are lots of issues related to this.

chrisarcand:
@europ Yes, that looks like an error specific to your platform; your OpenSSL is incompatible and you either need an updated OpenSSL or need to install ruby with specific flags to utilize it correctly; this will vary depending on your distro so I'm afraid we won't be a lot of help here.

skateman:
@chrisarcand is it safe for @europ to work with 2.4 (with the extra require) and push his changes into the 2.3 implementation (without the extra require)?

chrisarcand:
I mean, realize that what you're asking for is updating dependencies purely for the sake of a shortcoming on someone's local machine; you *should* be able to install Ruby 2.3 (the version we use for this project) on your laptop one way or another, and it's not miq-bot's concern at all here.

That said, if dependencies are updated that happen to work with both 2.3 and 2.4 and the tests pass here, I guess it's fine to do so. Note that you need to change the travis yml here to use 2.3.1 and tests must pass; we also need to ensure that there's no weird collision when you include xmlrpc with Ruby 2.3 (which already has it). Conditionally including it or not isn't acceptable as the lockfile will differ, so if it conflicts I'm afraid my opinion is just to say you need to fix your laptop to use Ruby 2.3.

PR #375, opened 2017-11-28
Add Repo#create_branch method

bdunne:
Given a name, it will check the git repo on the filesystem, and if the branch exists, fill in the details and create a Branch record

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -6,12 +6,10 @@ class Repo < ActiveRecord::Base
   validates :name, :presence => true, :uniqueness => true
 
   def self.create_from_github!(name, url)
-    create_and_clone!(name, url, Branch.github_commit_uri(name)).tap do |repo|
-      repo.ensure_prs_refs
-    end
+    create_and_clone!(name, url).tap(&:ensure_prs_refs)
   end
 
-  def self.create_and_clone!(name, url, commit_uri)
+  def self.create_and_clone!(name, url)
     path = BASE_PATH.join(name)
 
     raise ArgumentError, "a git repo already exists at #{path}" if path.join(".git").exist?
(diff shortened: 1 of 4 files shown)


app/models/repo.rb
```diff
@@ -149,6 +149,22 @@ def git_fetch
     end
   end
 
+  def create_branch(branch_name)
+    b = branches.new(
+      :name => branch_name,
+      :commit_uri => "https://github.com/#{branch_name}/commit/$commit",
```
Fryguy: Prefer `Branch.github_commit_uri`

app/models/repo.rb
```diff
@@ -149,6 +149,22 @@ def git_fetch
...
+    )
+
+    # Make sure the branch is a real git branch before continuing and saving a record
+    return if !b.git_service.send(:rugged_repo).branches.exists?("origin/#{b.name}")
```
Fryguy: I think there's a branch_exists? method on one of the git services (or if not, we should add it as I needed it for something else, if I recall)

app/models/repo.rb
```diff
@@ -149,6 +149,22 @@ def git_fetch
...
+
+    b.merge_target = "master"  # I know this shouldn't be, but I need it for the last_commit
+    b.last_commit  = b.git_service.merge_base
+    b.merge_target = nil
```
Fryguy: We have a similar method that creates the master branch in `Repo.create_from_github!`...can we DRY those up?

lib/git_service/branch.rb
```diff
@@ -22,12 +26,12 @@ def mergeable?
       false
     end
 
-    def merge_base
-      rugged_repo.merge_base(target_for_reference(merge_target_ref_name), target_for_reference(ref_name))
+    def merge_base(other = nil)
```
Fryguy: Instead of `other`, can this be `merge_target`...`other` is very cryptic to me.

PR #391, opened 2017-12-20
Update rubocop to 0.52.0

Fryguy:
Replaces #373 

@bdunne Please review

We will also have to update the yaml files in ManageIQ/guides because it seems a bunch of cops have had their categorization changed.


Fryguy:
Also I updated the cached results files to better reflect what's changed in 0.52.0, as well as making them pretty so that future diffs are clearer.

PR #398, opened 2018-01-17
Reduce callers to git_service

bdunne:
We no longer need to lock a repo while running the commit monitor.  Once the fetch is done, everything can be pulled from rugged.  This will unlock future performance improvements like processing multiple repos (and even branches) at the same time.

@Fryguy We need to do some testing before deploying, but everything was working for me in rails console.

### DIFF
--- a/app/workers/commit_monitor.rb
+++ b/app/workers/commit_monitor.rb
@@ -118,7 +118,7 @@ class CommitMonitor
     @new_commits_details ||=
       new_commits.each_with_object({}) do |commit, h|
         h[commit] = {
-          "message" => git.commit_message(commit),
+          "message" => branch.git_service.commit_message(commit),
           "files"   => git.diff_file_names(commit)
         }
       end
--- a/app/workers/commit_monitor.rb
+++ b/app/workers/commit_monitor.rb
@@ -55,23 +55,21 @@ class CommitMonitor
 
   private
 
-  attr_reader :repo, :git, :branch, :new_commits, :all_commits, :statistics
+  attr_reader :repo, :branch, :new_commits, :all_commits, :statistics
 
   def process_repo(repo)
     @statistics = {}
 
     @repo = repo
-    repo.with_git_service do |git|
-      @git = git
+    repo.git_fetch
 
-      # Sort PR branches after regular branches
-      sorted_branches = repo.branches.sort_by { |b| b.pull_request? ? 1 : -1 }
+    # Sort PR branches after regular branches
+    sorted_branches = repo.branches.sort_by { |b| b.pull_request? ? 1 : -1 }
 
-      sorted_branches.each do |branch|
-        @new_commits_details = nil
-        @branch = branch
-        process_branch
-      end
+    sorted_branches.each do |branch|
+      @new_commits_details = nil
+      @branch = branch
+      process_branch
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/git_service/branch.rb
```diff
@@ -65,6 +65,28 @@ def commit_ids_since(starting_point)
...
+      commit = Rugged::Commit.lookup(rugged_repo, ref)
+      parents = commit.parent_oids
+      message = "commit #{commit.oid}\n"
+      message << "Merge: #{parents.join(" ")}\n" if parents.length > 1
+      message << "Author:     #{commit.author[:name]} <#{commit.author[:email]}>\n"
+      message << "AuthorDate: #{commit.author[:time].to_time.strftime("%c %z")}>\n"
+      message << "Commit:     #{commit.author[:name]} <#{commit.author[:email]}>\n"
+      message << "CommitDate: #{commit.author[:time].to_time.strftime("%c %z")}>\n"
+      message << "\n"
+      commit.message.each_line { |line| message << "    #{line}"}
```
Fryguy: Prefer `message << commit.message.indent(4)`

lib/git_service/branch.rb
```diff
@@ -65,6 +65,28 @@ def commit_ids_since(starting_point)
...
+      message << "\n"
+      previous_commit = Rugged::Commit.lookup(rugged_repo, commit.parent_oids.first)
+      diff = previous_commit.diff(commit)
+      our_diff = Diff.new(diff)
+      our_diff.file_status.each do |file, stats|
+        message << " #{file} | #{stats[:changes]} #{"+" * stats[:additions]}#{"-" * stats[:deletions]}\n"
```
Fryguy: Need to deal with large number of changes...they do some kind of math where there is a max number of sigils, then everything is relative to that.

Also the file name is abbreviated in an interesting way if the name is too long.  It basically chops off prefix paths with `...` until it fits some common width

lib/git_service/diff.rb
```diff
@@ -24,5 +24,29 @@ def with_each_line
...
+      end
+    end
+
+    def status_summary
+      changed, added, deleted = raw_diff.stat
+      [].tap do |array|
+        array << "#{changed} #{"file".pluralize(changed)} changed" if changed > 0
+        array << "#{added} #{"insertion".pluralize(added)}(+)" if added > 0
+        array << "#{deleted} #{"deletion".pluralize(deleted)}(-)" if deleted > 0
+      end.join(", ")
```
Fryguy: You could also

```ruy
[
  changed > 0 ? part1 : nil,
  added > 0 ? part2 : nil
  ..
].compact.join(", ")

File: app/workers/commit_monitor.rb [hunk 4]
```diff
@@ -55,23 +55,21 @@ def process_repos
...
-  attr_reader :repo, :git, :branch, :new_commits, :all_commits, :statistics
+  attr_reader :repo, :branch, :new_commits, :all_commits, :statistics
 
   def process_repo(repo)
     @statistics = {}
 
     @repo = repo
-    repo.with_git_service do |git|
-      @git = git
+    repo.with_git_service { } # Still needed for fetching
```
Fryguy: Prefer `repo.git_fetch` to make this more explicit

Fryguy:
Really nice!!  Just a few comments to get through.

lib/git_service/branch.rb
bdunne: yeah... do you think we need to do this now?

PR #405, opened 2018-02-09
[WIP] Unclutter commit monitor

Fryguy:
This PR unclutters the core of the CommitMonitor by removing the whole categorization notion of separate workers.  All workers will receive the branch_id and the new_commits on that branch, and it is up to the worker how they want to deal with that.

@bdunne Please review.

PR is WIP because I haven't tested enough yet.


Fryguy:
First 2 commits are flattening the commit_range directory.  The last commit is the big change to the CommitMonitor

Fryguy:
Did a quick first pass on testing and something is wrong.  It seems to execute all of the handlers even if there haven't been any changes to the branch.  I must have removed a check somewhere in all those deletions.  No issues though when the actual workers run.

Fryguy:
Reopening cause I still think this is the right approacj

Issue #192, opened 2016-06-13
Review pull requests using pronto


skateman:
Hello! 

I've found [this gem](https://github.com/mmozuras/pronto), which is able to unify the output of multiple static code review services. I already have some contributions to the code and the maintainer was always helpful. This example shows how can a PR reviewed:

```ruby
#!/bin/env ruby
require 'fileutils'
require 'pronto'
require 'rugged'
require 'pronto/rubocop'

PR_N = 7470
BASE = 'd5b175f'.freeze
REPO = 'manageiq/manageiq'.freeze
TRGT = 'manageiq'.freeze

repo = Rugged::Repository.init_at("tmp/#{TRGT}/")
repo.remotes.create('origin', "https://github.com/#{REPO}.git")

pronto_repo = Pronto::Git::Repository.new("tmp/#{TRGT}")
rugged_repo = pronto_repo.instance_variable_get(:@repo)

rugged_repo.remotes['origin'].fetch("pull/#{PR_N}/head")
rugged_repo.checkout('FETCH_HEAD')

diff = rugged_repo.diff(BASE, rugged_repo.head.target)
patches = Pronto::Git::Patches.new(pronto_repo, BASE, diff)
result = Pronto::Runners.new.run(patches)
formatter = Pronto::Formatter::JsonFormatter.new
formatter.format(result, pronto_repo, patches)

FileUtils.rm_rf('tmp/manageiq')
```

I know that it's not the cleanest code, but it looks way better than our current one :innocent:

skateman:
cc @chriskacerguis

PR #406, opened 2018-02-13
Pronto integration

europ:
### Closes https://github.com/ManageIQ/miq_bot/issues/192

Pronto supports a [list of runners](https://github.com/prontolabs/pronto/blob/master/README.md#runners) available in [pronto/README.md](https://github.com/prontolabs/pronto/blob/master/README.md) which can be launched. To launch a runner, add the runner into [`Gemfile`](https://github.com/europ/miq_bot/blob/1fb4ca1d2fdc60efce9a3461d7014bc62419096c/Gemfile) ~~and require it properly in [code_analysis_mixin.rb](https://github.com/europ/miq_bot/blob/1fb4ca1d2fdc60efce9a3461d7014bc62419096c/app/workers/concerns/code_analysis_mixin.rb)~~. The proper requiring is [automated](https://github.com/ManageIQ/miq_bot/pull/406/commits/5383904b9fbfd8f30c7ab00c4844ee4531036460) and now it is necessary to specify only the pronto gem.

It was necessary due to integration of pronto to create a same data structure from the pronto output as the original was. The integration was done in file [`app/workers/concerns/code_analysis_mixin.rb`](https://github.com/ManageIQ/miq_bot/blob/master/app/workers/concerns/code_analysis_mixin.rb) in [`run_all_linters`](https://github.com/ManageIQ/miq_bot/blob/master/app/workers/concerns/code_analysis_mixin.rb#L22) method where is the code analysis launched via pronto and the result is formatted to match the originall pattern of the result structure.

Pronto result is an array of pronto messages which are iterated throught in order to create identical data structure of the former structure. The `pronto_result` method starts the pronto-runners a gather the informations and return an array of pronto messages. The `run_all_linters` method parses the array of pronto messages and creates a hash as the result. The hash structure contains all informations about the files inspected by pronto-runners which launch the corresponding linter.

---

* Due to some Pronto deficiencies few values are strictly defined, they were marked with `TODO`.
   **NOTE**: It was solved by removing theses variables (see the review comments below)

* Syntactic errors which are missing from the result of the Pronto-RuboCop will be solved by https://github.com/prontolabs/pronto-rubocop/pull/35

* Indention of the expected hash in `code_analysis_mixin_spec.rb` is extensive due to rubocop hash alignment correctness

* Informations about the position of the offence (column, line, length) are changed and from now line position is the only one supported because of request in comments below

---

\cc
@skateman 
@romanblanco


Fryguy:
@bdunne Please review.

app/workers/concerns/code_analysis_mixin.rb
```diff
@@ -19,11 +31,64 @@ def merged_linter_results
...
+
+    # temporary solution for: download repo, obtain changes, get pronto result about changes
+    Dir.mktmpdir do |dir|
+      FileUtils.copy_entry(@branch.repo.path.to_s, dir)
+      repo = Pronto::Git::Repository.new(dir)
+      rg = repo.instance_variable_get(:@repo)
+      rg.fetch('origin', @branch.name.sub(/^prs/, 'pull'))
+      rg.checkout('FETCH_HEAD')
+      rg.reset('HEAD', :hard)
+      patches = repo.diff(@branch.merge_target)
```
Fryguy: This is already possible with the repos from the repos directory.  Unfortunately, Pronto::Git::Repository seems to only want to use head, but it doesn't look like a stretch to just "copy" what they do there (or make a PR to pronto to support non-head commits, which should be straightforward).  I think you can just do something like

```ruby
branch.repo.git_fetch # note that branch is available to the CodeAnalysisMixin directly via the BranchWorkerMixin as they are usually together
git_service = branch.git_service
merge_base = git_service.merge_base(commit)
diff = git_service.rugged_repo.diff(merge_base, "refs/remotes/origin/#{branch.merge_target}", options)  # Getting the merge target ref is actually already a part of lib/git_service/branch.rb, but it's not exposed in an easy way, so just hardcoding here for an example
patches = Pronto::Git::Patches.new(Pronto::Git::Repository.new(branch.repo.path, merge_base, diff)
p_result = Pronto::Runners.new.run(patches)
```

app/workers/concerns/code_analysis_mixin.rb
```diff
@@ -19,11 +31,64 @@ def merged_linter_results
...
+              }
+            }
+          end
+        }
+      end
+
+      output["summary"] = {
+        "offense_count"        => output["files"].reduce(0) { |sum, item| sum + item['offenses'].length },
+        "target_file_count"    => output["files"].length,
+        "inspected_file_count" => 0 # TODO: value cannot be obtained from the result of `pronto_result` method
```
Fryguy: Do we even use this in the output comment?

app/workers/concerns/code_analysis_mixin.rb
```diff
@@ -19,11 +31,64 @@ def merged_linter_results
...
+                "column" => 0, # TODO: value cannot be obtained from Pronto::Message
+                "length" => msg.line.length
+              }
+            }
+          end
+        }
+      end
+
+      output["summary"] = {
+        "offense_count"        => output["files"].reduce(0) { |sum, item| sum + item['offenses'].length },
```
Fryguy: Alternately,

```
output["files"].sum { |item| item['offenses'].length }  # Though sum may only be available on Ruby 2.4 (not sure)
```

spec/workers/concerns/code_analysis_mixin_spec.rb
```diff
@@ -8,34 +8,188 @@
...
+      expect(subject.merged_linter_results).to eq("files" => [], "summary" => {"inspected_file_count" => 0, "offense_count" => 0, "target_file_count" => 0})
     end
+  end
 
-    it "merges together with one result" do
-      expect(Linter::Rubocop).to receive(:new).and_return(double(:run => example_rubocop_result))
-      expect(Linter::Haml).to receive(:new).and_return(double(:run => nil))
-      expect(Linter::Yaml).to receive(:new).and_return(double(:run => nil))
+  describe "#run_all_linters" do
+    let(:item_rubocop) { dup }
```
Fryguy: I've never seen `dup` like this before...what does this do?

Fryguy:
great PR @europ !!  I think this is a huge step forward.

> the result is formatted to match the originall pattern of the result structure.

This is great for now, but if pronto has their own format, perhaps we just drop that intermediate format altogether and just use the pronto format directly in the message builder.  The message builder was tightly intertwined with rubocop's output format, because at first that was the only thing we ran.  As we went forward we converted things like the Haml output into the rubocop format.  However if pronto already has a unified format, I think it's better to just use that directly. If we do choose to switch to the pronto format, let's do that in a separate PR.

Fryguy:
@europ Can you also delete all of the rubocop/linters directory since they are no longer used?

spec/workers/concerns/code_analysis_mixin_spec.rb
skateman: @Fryguy it's a dummy object for mocking

Fryguy: Is it the same as just calling `double`?

Fryguy:
One question...how do we configure rules for what pronto runs?  I know there's a pronto config, but do we need one in every target repo.  If so, how can we handle some of the rubocop yml file based configuration we do (which uses shared config)

spec/workers/concerns/code_analysis_mixin_spec.rb
Fryguy: I don't think this is doing what you think it does  (I can't find any docs on RSpec providing a dup method).  If anything it's probably calling Object#dup on the current context?  I think I would prefer `double` here, so that it's explicit that we want a mock object.

Fryguy:
@europ Can you also put somewhere in the commit message that this `Closes #192`

skateman:
@Fryguy AFAIK it needs to have a configfile for each runner (i.e. rubocop, hamllint, yamllint) in every target repo. But I like your idea about the global config and I don't think this would be a problem to implement, we could automatically pull the linter configurations from a single place while running. However, this way we would not have the linter configurations in local repo forks.

Fryguy:
@europ Can you please remove my name from the commit messages?  I get an email every time you rebase :)

app/workers/concerns/code_analysis_mixin.rb
europ: @Fryguy it is not used, it was used only in /lib/linters, the output comment does not use it https://github.com/ManageIQ/miq_bot/blob/master/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb#L44
I will remove it.

europ:
@miq-bot remove_label unmergeable

NickLaMuro:
So looking into updating this, as it pushes us into the direction of getting https://github.com/ManageIQ/miq_bot/issues/499 moving forward:

Aside from this conflicting with master, this also is having issues with the branch that I expect to be merged soon:

https://github.com/ManageIQ/miq_bot/pull/498

There are a few weird down grades we are taking with dependent gems as a result:

```
Note: rainbow version regressed from 3.0.0 to 2.2.2
Note: thor version regressed from 1.0.1 to 0.19.4
```

(the above from the `bundler` output).

This is because of the pronto changes.  Shouldn't be a big deal, but just a heads up.  Currently trying to fix specs since regenerating the `results.json` files is not going smoothly at the moment...

Fryguy:
Closing in favor of #500

PR #409, opened 2018-03-01
Declare dependency on xmlrpc

bdunne:
Allow development with ruby 2.5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,3 +66,5 @@ group :test do
   gem 'webmock'
   gem 'factory_girl_rails'
 end
+
+gem "xmlrpc"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -354,6 +354,7 @@ GEM
       crack (>= 0.3.2)
       hashdiff
     websocket (1.2.4)
+    xmlrpc (0.3.0)
 
 PLATFORMS
   ruby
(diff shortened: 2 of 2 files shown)


chrisarcand:
If we do this, the lockfile will change on some systems but not others causing a dirty change. We currently check in the lockfile, as should be the case when you're developing on an application (as opposed to a gem).

As we can't use 2.5 features anyway because it needs to be compatible with 2.3 for ManageIQ, I don't know why we'd want to do this - but if you really really want to, would it conflict if  you added it regardless of it being needed for < 2.5 (to keep the lockfiles the same in both cases...?)

chrisarcand:
@bdunne Done any trying out of this on < 2.5 to make sure the gem doesn't screw with the existing lib?

bdunne:
Nope.  I can close this if we're not ready yet (since it will be quite a while before we're on 2.5) or mark as WIP and collect any other 2.5 issues in this PR as well

chrisarcand:
¯\\\_(ツ)_/¯ Doesn't matter to me! I'm cool with whatever, as long as the lockfile is still able to keep env parity as it's designed to do and the lib doesn't screw with the core one.

bdunne:
@miq-bot add_reviewer @Fryguy

bdunne:
@miq-bot add_reviewer @Fryguy

juliancheal:
@bdunne This is fixed by #430

Fryguy:
Closed via #434

PR #408, opened 2018-03-01
Review request command

europ:
### Related issue: https://github.com/ManageIQ/miq_bot/issues/337

ManageIQ Bot will support a new command which provides a "request for pull request review" functionality. This feature required to update Octokit due to [`request_pull_request_review`](https://octokit.github.io/octokit.rb/Octokit/Client/Reviews.html#request_pull_request_review-instance_method) method missing. This command works on the same basis as the [`assign`](https://github.com/ManageIQ/miq_bot) command but it is supported only for pull requests.

Example: `@miq-bot review @europ`

**NOTE**: Command with multiple reviewers are not supported (assign does not support this feature too).

\cc @skateman 
\cc @romanblanco

PR #411, opened 2018-03-06
Remove a request for pull request review command

europ:
### Dependent on: https://github.com/octokit/octokit.rb/pull/990

<br>

ManageIQ Bot will support a new command `remove_reviewer`. This command provides a "remove request for pull request review" functionality. The usage of this command requires to meet a condition - the user must be a requested reviewer in the pull request. If condition is met, the command perform an action which removes the user from the reviewer list in the pull request.  This command is the opposite of `add_reviewer` which is described in https://github.com/ManageIQ/miq_bot/pull/408

Example: `@miq-bot remove_reviewer @europ`

**NOTE**: Command with multiple reviewers is not supported

\cc @skateman 
\cc @romanblanco

### DIFF
--- a/spec/lib/github_service/commands/remove_reviewer_spec.rb
+++ b/spec/lib/github_service/commands/remove_reviewer_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+
+RSpec.describe GithubService::Commands::RemoveReviewer do
+  subject { described_class.new(issue) }
+  let(:issue) { double(:fq_repo_name => "org/repo") }
+  let(:command_issuer) { "nickname" }
+
+  before do
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "listed_user").and_return(true)
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "good_user").and_return(true)
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "bad_user").and_return(false)
+
+    allow(subject).to receive(:requested_reviewers).and_return(["listed_user"])
+  end
+
+  after do
+    subject.execute!(:issuer => command_issuer, :value => command_value)
+  end
+
+  context "with a valid user who is actually requested for a review" do
+    let(:command_value) { "listed_user" }
+
+    it "remove review request from that user" do
+      expect(issue).to receive(:remove_reviewer).with("listed_user")
+    end
+  end
+
+  context "with a valid user who is not actually requested for a review" do
+    let(:command_value) { "good_user" }
+
+    it "does not remove review request who is not actually requested" do
+      expect(issue).not_to receive(:remove_reviewer)
+      expect(issue).to receive(:add_comment).with("@#{command_issuer} 'good_user' is not in the list of requested reviewers, ignoring...")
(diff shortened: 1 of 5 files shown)


Fryguy:
LGTM!  Please add a spec as well.

spec/lib/github_service/commands/remove_reviewer_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+RSpec.describe GithubService::Commands::RemoveReviewer do
+  subject { described_class.new(issue) }
+  let(:issue) { double(:fq_repo_name => "org/repo") }
+  let(:command_issuer) { "nickname" }
+
+  before do
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "listed_user") { true }
```
skateman: @europ what does the `{ true }` block at the end do here?

spec/lib/github_service/commands/remove_reviewer_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "good_user") { true }
+    allow(GithubService).to receive(:valid_assignee?).with("org/repo", "bad_user") { false }
+
+    allow(subject).to receive(:requested_reviewers).and_return(["listed_user"])
+  end
+
+  after do
+    subject.execute!(:issuer => command_issuer, :value => command_value)
+  end
```
skateman: Why is this necessary? What are you cleaning up here?

spec/lib/github_service/commands/remove_reviewer_spec.rb
europ: @skateman .and_return(true)

spec/lib/github_service/commands/remove_reviewer_spec.rb
europ: It is used also in [assign_spec.rb](https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/github_service/commands/assign_spec.rb#L14), [add_reviewer_spec.rb](https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/github_service/commands/add_reviewer_spec.rb#L14), [add_label_spec](https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/github_service/commands/add_label_spec.rb#L18), [remove_label_spec.rb](https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/github_service/commands/remove_label_spec.rb#L10). I took inspiration from these mentioned specs before and on this basis I created my remove_reviewer_spec.rb.

Fryguy: It's weird, but all of the others do their thing-under-test in an after block.  I don't like it, but for this PR, it's consistent.

europ:
@Fryguy [this](https://github.com/ManageIQ/miq_bot/pull/411/files#diff-7c743e8ae1171ab8956f1b6d6b4ff439R18) will not work without https://github.com/octokit/octokit.rb/pull/990 which is not merged yet.

PR #412, opened 2018-03-07
Integration of GitHub Status API

europ:
### Description: https://developer.github.com/v3/repos/statuses/#create-a-status

Commits are marked with `success` or `error` state depending on offence(s) detection. If there are any offences in the code, the commit is marked with `error` state, otherwise with `success` state. It is reflected at the end of pull request in the `checks` section. The status also involves a reference to the comment which has the description about the offences.

\cc
@skateman 
@romanblanco

---

### Preview

* #### success

good_status

* #### error

bad_status


app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -34,12 +34,14 @@ def process_branch
   end
 
   def rubocop_comments
-    MessageBuilder.new(results, branch).comments
+    message_builder = MessageBuilder.new(results, branch)
+    @status = message_builder.commit_status
```
skateman: I don't think this is *The Right Way&trade;* to pass the information, the `rubocop_comments` method is being used just in the `replace_rubocop_comments` method, so it's safe to move its content there. This way you don't need the instance variable at all. Then you can refactor or even rename the method and maybe break it into small(er) chunk.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -34,12 +34,14 @@ def process_branch
...
+    message_builder.comments
   end
 
   def replace_rubocop_comments
     logger.info("Updating PR #{pr_number} with rubocop comment.")
-    GithubService.replace_comments(fq_repo_name, pr_number, rubocop_comments) do |old_comment|
+    GithubService.replace_comments(fq_repo_name, pr_number, rubocop_comments, @status) do |old_comment|
```
skateman: See my comment above...

lib/github_service.rb
```diff
@@ -11,12 +11,22 @@ module GithubService
...
   class << self
-    def add_comments(fq_repo_name, issue_number, comments)
+    def add_comments(fq_repo_name, issue_number, comments, commit_status = nil)
       Array(comments).each do |comment|
-        add_comment(fq_repo_name, issue_number, comment)
+        retval = add_comment(fq_repo_name, issue_number, comment)
+
+        unless commit_status.nil?
+          commit_status["options"]["target_url"] = retval["html_url"]
+          add_status(commit_status)
```
skateman: Adding the status is depending on the comment's ID, however, it is an operation on the same level so this method call doesn't belong here, but AFTER the `add_comments` is called.

lib/github_service.rb
```diff
@@ -25,12 +35,12 @@ def delete_comments(fq_repo_name, comment_ids)
...
     # Deletes the issue comments found by the provided block, then creates new
     # issue comments from those provided.
-    def replace_comments(fq_repo_name, issue_number, new_comments)
+    def replace_comments(fq_repo_name, issue_number, new_comments, commit_status = nil)
       raise "no block given" unless block_given?
 
       to_delete = issue_comments(fq_repo_name, issue_number).select { |c| yield c }
       delete_comments(fq_repo_name, to_delete.map(&:id))
-      add_comments(fq_repo_name, issue_number, new_comments)
+      add_comments(fq_repo_name, issue_number, new_comments, commit_status)
```
skateman: I think you should put the `add_status` call after this line, see above...

Gemfile.lock
```diff
@@ -396,4 +396,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.15.4
+   1.16.1
```
skateman: Don't forget to change this back please...

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -34,12 +34,19 @@ def process_branch
...
+      :comments => message_builder.comments
+    }
   end
 
   def replace_rubocop_comments
     logger.info("Updating PR #{pr_number} with rubocop comment.")
-    GithubService.replace_comments(fq_repo_name, pr_number, rubocop_comments) do |old_comment|
+    msg_builder_result = rubocop_comments
+
+    GithubService.replace_comments(fq_repo_name, pr_number, msg_builder_result[:comments], msg_builder_result[:status]) do |old_comment|
```
skateman: Do you really need the 2 arguments if you're taking them from the same hash? Send the whole hash instead!

app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
```diff
@@ -14,6 +14,20 @@ def comments
     message_builder.comments
   end
 
+  # requirements to "create_status(repo, sha, state, options)"
+  def commit_status
```
skateman: Why is this method under the `MessageBuilder`? Maybe this can be moved to a better place.

lib/github_service.rb
```diff
@@ -25,12 +34,15 @@ def delete_comments(fq_repo_name, comment_ids)
...
+    def replace_comments(fq_repo_name, issue_number, new_comments, commit_status = nil)
       raise "no block given" unless block_given?
 
       to_delete = issue_comments(fq_repo_name, issue_number).select { |c| yield c }
       delete_comments(fq_repo_name, to_delete.map(&:id))
-      add_comments(fq_repo_name, issue_number, new_comments)
+      first_comment = add_comments(fq_repo_name, issue_number, new_comments)
+
+      # add_status creates a commit status pointing to the first comment URL
+      add_status(commit_status, first_comment)
```
skateman:
```ruby
add_status(commit_status, first_comment) unless first_comment.nil?
```
Then you don't need the condition in the `add_status`.

lib/github_service.rb
```diff
@@ -12,8 +12,17 @@ module GithubService
...
+      Array(comments).map do |comment|
         add_comment(fq_repo_name, issue_number, comment)
+      end.first
+    end
+
+    # https://octokit.github.io/octokit.rb/Octokit/Client/Statuses.html#create_status-instance_method
+    def add_status(comstat, comment)
+      unless comstat.nil?
+        comstat["options"]["target_url"] = comment["html_url"]
+        create_status(comstat["repo"], comstat["sha"], comstat["state"], comstat["options"])
```
skateman: Again, you're sending arguments from a single hash, these things can be simplified.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
Fryguy: Agreed.  This feels like it should live somewhere else.

lib/github_service.rb
```diff
@@ -12,9 +12,16 @@ module GithubService
   #
   class << self
     def add_comments(fq_repo_name, issue_number, comments)
-      Array(comments).each do |comment|
+      # add_status requires first comment due to its URL
+      Array(comments).map do |comment|
         add_comment(fq_repo_name, issue_number, comment)
-      end
+      end.first
```
Fryguy: I like the idea of returning the comments, but from a clean standpoint, this method should just return all of the comments.  If the caller only needs the first one, they should be the one to call .first.  This way you also don't need the comment describing why you need the .first.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb
europ: @skateman @Fryguy The `commit_status` method is now in the `MessageBuilder`.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -33,13 +33,31 @@ def process_branch
...
+  # requirements to "create_status(repo, sha, state, options)"
+  # state is a boolean variable describing success/error state
+  def commit_status(state)
+    {
+      "repo"    => fq_repo_name,
+      "sha"     => commits.last,
+      "state"   => state ? "success" : "error",
+      "options" => {
+        "context"     => "miq-bot", # TODO: it should be user name variable
+        "target_url"  => nil,
```
skateman: Why is this `nil`? Can't be this parametrized as `state`?

europ: @skateman The key is required and the value is added when the URL of the comment is obtained.

skateman: OK, what about the 2nd part of the question? :wink:

skateman:
@europ please don't mark this as ready until you have tests.

@miq-bot add_label wip

europ:
@miq-bot remove_label wip

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
```diff
@@ -39,7 +39,10 @@ def rubocop_comments
 
   def replace_rubocop_comments
     logger.info("Updating PR #{pr_number} with rubocop comment.")
-    GithubService.replace_comments(fq_repo_name, pr_number, rubocop_comments) do |old_comment|
+
+    status = results["files"].select { |f| f["offenses"].any? }.empty?
```
Fryguy: Can you just do `status = results["files"].any? { |f| f["offenses"].any? }` ?

spec/lib/github_service_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+
+RSpec.describe GithubService do
+  let(:repo) { "owner/repository" }
+  let(:sha) { "cd16q9z48b3b0xbb91wb311" }
```
Fryguy: Pretty sure a SHA only has hexidecimal digits ;)

Fryguy:
high level I like the ability to use the github api status, but I'm not sure I want to block the PR expect in specific cases (like :bomb: :boom: :fire: :fire_engine:)

skateman:
@fryguy by block do you mean red status? AFAIK we don't block PRs based on any status...

Fryguy:
Sorry, yes, red status.  If too many PRs go "red" then it's harder for reviewers to differentiate between actual failed PRs (i.e. travis didn't pass) vs not actually failed PRs (like how when code climate goes red on 1 new issue).

lib/github_service.rb
```diff
@@ -25,12 +39,17 @@ def delete_comments(fq_repo_name, comment_ids)
...
+    def replace_comments(fq_repo_name, issue_number, new_comments, status = nil, commit_sha = nil)
       raise "no block given" unless block_given?
 
       to_delete = issue_comments(fq_repo_name, issue_number).select { |c| yield c }
       delete_comments(fq_repo_name, to_delete.map(&:id))
-      add_comments(fq_repo_name, issue_number, new_comments)
+      first_comment = add_comments(fq_repo_name, issue_number, new_comments).first
+
+      # add_status creates a commit status pointing to the first comment URL
+      unless first_comment.nil? && status.nil?
```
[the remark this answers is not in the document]
europ: @Fryguy Can it be like this?
```ruby
if first_comment && !status.nil?
  add_status(...)
end
```
The status variable contain a boolean value so your example will not work in case of status has `False` value.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
europ: It works exactly the opposite. I will have to negate it as
```ruby
status = !results["files"].any? { |f| f["offenses"].any? }
```
otherwise I will have to remake the rest of the code. Do you agree with the negation?

skateman:
@Fryguy this makes sense, so let's add a green status when all is good and a red when :bomb: :boom: :fire: :fire_engine: happens. For other cases I would just simply not set the status...what do you think?

lib/github_service.rb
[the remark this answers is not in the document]
Fryguy: Ah ok, then yes, `first_comment && !status.nil?` is good.

app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb
Fryguy: if you're looking for the negation, then you can also do

```ruby
results["files"].none? { |f| f["offenses"].any? }
```

europ:
Based on the pull request review from @skateman and @Fryguy this feature was a little bit changed.

### There are now 3 types of status:

* **Green status** (zero offences)<br>
status_green

* **Red status** (at least on fatal or error offence)<br>
status_red

* **No status** (other offence types such as warning, informing, etc.)<br>
status_none

Fryguy:
@europ I'm thinking the third case

> No status (other offense types such as warning, informing, etc.)

should still show the status API entry, but it should be a) green and b) say something like "Changes requested"

europ:
@Fryguy There is a possibility to use "yellow" status - the pending status.

There are 4 available statuses:
* **failure** - this means an internal error (red) - this will not be used
* **error** - at leas one :bomb: :boom: :fire: :fire_engine: offense (red)
* **pending** - detected offenses required to fix (yellow)
* **success** - no offense detected (green)

Would you like to use the "yellow" one (pending status) for this case with a message "*detected offenses required to fix*"?

Fryguy:
The problem there is that when PR reviewers see "yellow" they think "the CI for this PR is still running", which is not what we want.  I'm thinking just use green.

europ:
@Fryguy The commits are from now marked every time depending of the detected offences as you requested.

There are 3 types of commit marking:
1. **zero** offences:
    "success" status + "Everything looks fine."
2. **one or more** warning offences excluding :bomb: :boom: :fire: :fire_engine:  offences:
    "success" status + "Some offenses detected."
3. **at least one** :bomb: :boom: :fire: :fire_engine:  offence:
    "error" status + "Something went wrong."

PR #417, opened 2018-03-27
[WIP] Automated request of codeowner for PR review

europ:
### **[WIP]**

This was necessary due to the fact that there is not possible to choose number zero for required approving reviews in the settings of branch protection. Codeowners are requested for a pull request review depending on specifications defined in CODEOWNERS file. This file includes files specifications and authors related to them. Files from the PR are compared if they match any of the pattern defined in the codeowners file. If there is match then a corresponding user is requested for a review.

### DIFF
--- a/app/workers/codeowner_review_request.rb
+++ b/app/workers/codeowner_review_request.rb
@@ -1,0 +1,165 @@
+require 'pry'
+require 'pry-byebug'
+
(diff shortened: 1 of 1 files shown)


europ:
@skateman 

GitHub has fixed this already. This pull request is useless from now ...

PR #418, opened 2018-04-04
Added missing linter into README.md

europ:
GitHub pull requests are also checked with yamllint [here](https://github.com/ManageIQ/miq_bot/blob/master/app/workers/concerns/code_analysis_mixin.rb#L26). Yamllint is also a miq_bot's dependency defined in the [setup](https://github.com/ManageIQ/miq_bot#setup).

image


juliancheal:
LGTM 👍  @bdunne is this good to merge?

PR #426, opened 2018-10-24
Add (suggestion for) better regexp for puts

NickLaMuro:
This is less of a "new feature" and more of a way we should consider handling checking for `puts` in our `config/settings.yml`.

In the current config by just looking for `puts` as a "word", we will end up flagging it when it is in comments or just in an everyday sentence (you know... since it is a 'word' in the English dictionary).

The suggestion I am proposing basically is to make the `puts` check a regexp that will not match against any line that includes a `#` proceeding the word, and I re-worked the specs to show how that would work.

* * *

The alternative to this approach would be to add the following regexp matcher preamble:

```ruby
/^([^#]+|)\b#{my_bad_word}\b/
```

To the "word" `Regexp` for of the content checker.  I figured this was more invasive so I went with the above, but might be more flexible overall and something to consider.

### DIFF
--- a/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
+++ b/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
@@ -4,7 +4,8 @@ describe CommitMonitorHandlers::CommitRange::GithubPrCommenter::DiffContentCheck
   let(:batch_entry)        { BatchEntry.create!(:job => BatchJob.create!) }
   let(:branch)             { create(:pr_branch) }
   let(:content_1)          { "def a(variable)" }
-  let(:content_2)          { "puts 'hi'" }
+  let(:content_2)          { "# this is going to `puts` 'hi'" }
+  let(:content_3)          { "puts 'hi'" }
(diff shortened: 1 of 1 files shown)


NickLaMuro:
> Some comments on commit NickLaMuro@d094e70
>
> **spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb**
>
> * [ ] ⚠️ - 44 - Detected puts. Remove all debugging statements.
> * [ ] ⚠️ - 7 - Detected puts. Remove all debugging statements.
> ...

Oh, thanks @miq-bot for showing me _**exactly**_ what was infuriating about this feature in the first place...

😡

juliancheal:
LGTM 👍 can you fix the `puts` errors above? :trollface:

NickLaMuro:
> LGTM 👍 can you fix the puts errors above? :trollface:

@juliancheal (and I guess a @Fryguy FYI):  This ends up requiring a config change as well to do it, so to actually remove this, we would need to update the regexp in the config to what I have in the specs.

Fryguy:
Yup..working on it

NickLaMuro:
Thanks, and just making sure... my PR descriptions in the past have known to cause the "glazing of the eyes"

PR #434, opened 2019-02-12
Remove active_bugzilla

NickLaMuro:
Replaced with some inline http calls via Faraday.  Uses the REST API, since the XMLRPC API has been deprecated and removed in BZ 5.0, and the bits needed for this app are small enough to not warrant trying to get all of `active_bugzilla` up to to code with the new API format.

Mostly marked [WIP] at this point because I haven't tested the PUT/POST actions.

Links
-----

* API docs specific to RedHat Bugzilla:
  - https://bugzilla.redhat.com/docs/en/html/api/core/v1/general.html (for authentication)
  - https://bugzilla.redhat.com/docs/en/html/api/core/v1/bug.html (GET/PUT bugs)
  - https://bugzilla.redhat.com/docs/en/html/api/core/v1/comment.html (GET/CREATE comments)

QA
--

Just a command line way to test the `GET` requests in the `BugzillaService`

1. Setup your bugzilla settings in `config/settings/development.yml`:
   
   ```yaml
   bugzilla_credentials:
     url: https://bugzilla.redhat.com
     username: "YOUR_USERNAME"
     password: "YOUR_PASSWORD"
   bugzilla:
   product: "Red Hat CloudForms Management Engine"
   ```
   
2. Run the following in the app dir with the branch checked out (make sure to `bundle` and such, and update the `1234567` with an ID of a valid BZ ID)
   
   ```console
   $ bin/rails runner "BugzillaService.call {|service| puts service.find_bug(1234567).inspect }" # fetch a bug
   $ bin/rails runner "BugzillaService.call {|service| puts service.find_bug(1234567).comments.inspect }" # fetch comments for that bug
   ```

### DIFF
--- a/app/workers/commit_monitor_handlers/commit_range/bugzilla_pr_checker.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/bugzilla_pr_checker.rb
@@ -32,7 +32,6 @@ module CommitMonitorHandlers
 
             add_pr_comment(bug)
             update_bug_status(bug)
-            bug.save
           end
         end
       end
@@ -53,8 +52,8 @@ module CommitMonitorHandlers
       end
 
       def bug_has_pr_uri_comment?(bug)
-        bug.comments.any? do |c|
-          c.text.include?(@branch.github_pr_uri)
+        bug.comments.any? do |comment_text|
+          comment_text.include?(@branch.github_pr_uri)
         end
       end
 
@@ -63,6 +62,7 @@ module CommitMonitorHandlers
         when "NEW", "ASSIGNED"
           logger.info "Changing status of bug #{bug.id} to ON_DEV."
           bug.status = "ON_DEV"
+          bug.save
         else
           logger.info "Not changing status of bug #{bug.id} due to status of #{bug.status}"
         end
--- a/lib/bugzilla_service.rb
+++ b/lib/bugzilla_service.rb
@@ -1,5 +1,5 @@
 class BugzillaService
-  include ThreadsafeServiceMixin
+  include ServiceMixin
 
   CLOSING_KEYWORDS = %w(
     close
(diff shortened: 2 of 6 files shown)


app/workers/commit_monitor_handlers/commit_range/bugzilla_pr_checker.rb
```diff
@@ -32,7 +32,6 @@ def update_bugzilla_status(bug_id)
 
             add_pr_comment(bug)
             update_bug_status(bug)
-            bug.save
```
Fryguy: The reason this was here was so that the PR comment and bug status go together as a single BZ POST (not sure that actually happened, but that was the intent.  I see this moves the save down to the update_bug_status.  Is there a similar save in add_pr_comment?

lib/bugzilla_service.rb
```diff
@@ -13,26 +13,126 @@ class BugzillaService
...
+    # Adds a comment.  Returns true if status == 200
+    def add_comment(text)
+      payload  = {"comment" => text}.to_json
+      response = service.post("/rest/bug/#{id}/comment", payload)
+
+      response.status == 200
+    end
+
+    # Basicially API compatible with `active_bugzilla` way of updating the bug
+    # status.
```
Fryguy: Oh I see, so with the REST API it *can't* be done in a single call?

Fryguy:
This is awesome!  So far, LGTM :+1:.  Can't wait for it to un-WIP.

app/workers/commit_monitor_handlers/commit_range/bugzilla_pr_checker.rb
NickLaMuro: Also answering https://github.com/ManageIQ/miq_bot/pull/434#discussion_r256490515 as well:

> Oh I see, so with the REST API it _can't_ be done in a single call?

So neither the REST API or the XMLRPC API was able to do this in a single call, but `active_bugzilla` has `active_model` included, so it just does a dirty state check on it's attributes, and if there is no attributes changed, it just no-ops.

https://github.com/ManageIQ/active_bugzilla/blob/3672da87bc25ce057978cf8fed9688670c09a50a/lib/active_bugzilla/bug.rb#L34-L39

Since we already do a check if we are going to do an update down in `update_bug_status`, I just moved the `bug.save` down to there since that is the only place we actually call a `.save`.

lib/bugzilla_service.rb
NickLaMuro: Answered above:  https://github.com/ManageIQ/miq_bot/pull/434#discussion_r256509574

juliancheal:
This is great @NickLaMuro 👍

juliancheal:
@NickLaMuro What's left to do to get the un-whipped?

NickLaMuro:
@juliancheal this:

> Mostly marked [WIP] at this point because I haven't tested the PUT/POST actions.

I don't want to merge this without confirming that I can do the following:

```console
$ bin/rails runner "BugzillaService.call {|service| bug = service.find_bug(1234567); bug.status = 'POST'; bug.save }" # update a bug
$ bin/rails runner "BugzillaService.call {|service| bug = service.find_bug(1234567); bug.add_comment 'Fixed by https://github.com/ManageIQ/manageiq/pull/123456' }" # add comment
```

NickLaMuro:
Alright, I actually got a chance to update this, and think we are in a good spot:

I updated a status and comment here:  https://bugzilla.redhat.com/show_bug.cgi?id=1648412#c36

Had to make the comment private after the fact (mildly sensitive data in it that I noticed "post posting"), which isn't relevant for the bot use cases I think.  Just ran the following:

```console
$ bin/rails runner "BugzillaService.call {|service| bug = service.find_bug(1648412); bug.status = 'ON_DEV'; bug.save }"
$ bin/rails runner "BugzillaService.call {|service| bug = service.find_bug(1648412); bug.add_comment File.read('bz_comment.txt') }"
```

(FYI:  'bz_comment.txt' was just a file where I pre-wrote and formatted the text for that comment)

@Fryguy @juliancheal think this is ready for review/merge now

juliancheal:
👍 LGTM @Fryguy

lib/bugzilla_service.rb
```diff
@@ -13,26 +13,126 @@ class BugzillaService
...
+      end
     end
   end
 
   def find_bug(id)
-    ActiveBugzilla::Bug.find(:product => Settings.bugzilla.product, :id => id).first
+    response = get("/rest/bug") do |req|
+      product = Settings.bugzilla.product
+      req.params["id"]             = id
+      req.params["product"]        = product if product
```
Fryguy: This is required, so I think we need to ensure product exists otherwise raise an exception.  We can't accidentally mess with bugs that aren't "ours"

lib/bugzilla_service.rb
```diff
@@ -13,26 +13,126 @@ class BugzillaService
...
+    end
   end
 
   def with_bug(id)
     yield find_bug(id)
   end
 
+  # Shared request helpers for adding BZ auth to a request
+  def get(path, &block)
+    connection.get(&faraday_request_block(path, &block))
```
Fryguy: I LOVE this...it's so nice.

lib/bugzilla_service.rb
```diff
@@ -13,26 +13,126 @@ class BugzillaService
...
+  # and POST requests)
+  def faraday_request_block(path)
+    credentials = [
+      Settings.bugzilla_credentials.username,
+      Settings.bugzilla_credentials.password
+    ].compact
+
+    lambda do |req|
+      req.url(path)
+      unless credentials.empty?
```
Fryguy: We probably should check that both username and password are set, else blow up, a la https://github.com/ManageIQ/active_bugzilla/blob/master/lib/active_bugzilla/service.rb#L66

Fryguy: One other note on accessing Settings directly, is that there is a service level accessor for this which is initialized at boot time.

https://github.com/ManageIQ/miq_bot/blob/5d1537dfb8288f4ed96b7b0907fe73fe836f4988/lib/service_mixin.rb#L11-L14
https://github.com/ManageIQ/miq_bot/blob/5d1537dfb8288f4ed96b7b0907fe73fe836f4988/config/initializers/services_initializer.rb#L2

The original intent was that a service could override the setter, and raise an exception is something was missing, but it seems BugzillaService never had that, instead relying on ActiveBugzilla to do the check.

Fryguy:
Just noticed something around ensuring parameter settings and how we access parameters.

Also, I *think* your implementation doesn't necessarily keep a service object instance around (unlike the other services).  So, I *think* you can remove the ThreadsafeServiceMixin as well?  If so that would be *awesome*, because then we don't have to serialize Bugzilla calls anymore.

NickLaMuro:
> So, I _think_ you can remove the `ThreadsafeServiceMixin` as well? If so that would be _awesome_, because then we don't have to serialize Bugzilla calls anymore.

Oh, I didn't even realize that was the purpose of that mixin.  Yeah, that should be doable, if not I can make it so.

lib/bugzilla_service.rb
NickLaMuro: > One other note on accessing Settings directly, is that there is a service-class level accessor for this which is initialized at boot time.

Yeah, I had a bit of trouble getting that to work, but I think that makes a bit more sense on how that works now.  Will look into that.

lib/bugzilla_service.rb
NickLaMuro: Roger, I will make it so.

Make it so

Fryguy:
yeah, the purpose of ThreadsafeServiceMixin is to serialize services that can't run simultaneously.  Since we are using Sidekiq, which uses threads, we have to be careful.  The un-threadsafe-ness is usually because either

- the service client object itself is not threadsafe (e.g. ActiveBugzilla used class-vars and was not threadsafe)
- something underlying is not thread safe (e.g. git itself won't be happy if we try to mess with commits at the same time as rebasing)
- we're not sure if the service client object is threadsafe or not, and we are just being careful

NickLaMuro:
Yeah, we seem to just be using it for the git service (valid), and bugzilla, which after this PR, isn't really necessary.

NickLaMuro:
@Fryguy Gotta re-run some of the QA stuff now, but I think this should address everything you had.

Left the last two commits as [FIXUP] since I plan on rebasing them after you review, but thought it made it easier to see the changes.  One change I have already amended into the first commit was the swapping of `ServiceMixin` in place of `ThreadsafeServiceMixin`, which was two lines of change.  Let me know if I need to point that one out to you, but should be pretty obvious.

NickLaMuro:
Added another comment in the above BZ, and it worked fine, so I think the updates are still valid.

NickLaMuro:
@Fryguy Done.

Issue #427, opened 2018-10-25
`miq_bot` update issues/checklist


NickLaMuro:
I am opening this issue to make note of and create a checklist for a number of issues with `miq_bot` when working on newer versions of Ruby/Rails:

- [x] `rails` is out of date (causes deprecation warnings with the `pg` gem)  ~~https://github.com/ManageIQ/miq_bot/pull/430~~ ~~https://github.com/ManageIQ/miq_bot/pull/435~~ https://github.com/ManageIQ/miq_bot/pull/466
- [x] `active_bugzilla` doesn't work on Ruby 2.4+ (`xmlrpc` was removed from stdlib)  https://github.com/ManageIQ/miq_bot/pull/466
  - Note:  This should really be addressed from `active_bugzilla` which there is a lot that needs to be updated there as well
  - Update:  Also, turns out Bugzilla dropped support for xmlrpc, so we now need to support the JSON API in that gem as well.
  - Update 2: As [mentioned below](https://github.com/ManageIQ/miq_bot/issues/427#issuecomment-463027982), this was addressed by simply removing the need for `xmlrpc`.

Going to try and address these myself, and might add to this list, but if you see PRs poping up from me, that is probably why.

juliancheal:
@NickLaMuro hey Nick, these are both fixed with #430

NickLaMuro:
@juliancheal This just fixes the first one.  The second one is unrelated to your fix.

NickLaMuro:
> The second one is unrelated to your fix.

Looks like I need to clarify:

The update to `xmlrpc` will allow the `miq_bot` project to bundle, but it will completely brick the `active_bugzilla` gem since it they removed the parser that we switched to in https://github.com/ManageIQ/active_bugzilla/pull/49 as  it was removed when it was still in ruby core:

- `XMLTreeParser` remove from `xmlrpc` (which worked with `xmlparser`): ruby/ruby@d8e3a92
- `xmlrpc` moved to a gem: https://bugs.ruby-lang.org/issues/12160

(ripped from a further discussion I had in the [bot channel](https://gitter.im/ManageIQ/miq_bot?at=5bf593f3d24f9324d22ff46d))

We don't have tests around that, since we stub it out, but I was attempting to address that here:

https://github.com/ManageIQ/active_bugzilla/pull/52

[...]

PR #435, opened 2019-02-15
Try rails 5 2 & Ruby 2.5.3

juliancheal:


juliancheal:
@miq-bot remove_label unmergeable

Fryguy:
Can we bump rubocop in a separate PR?  I'm concerned about mixing too many things in one PR, particularly if we have to undo anything.

juliancheal:
@Fryguy ah yeah good point.

juliancheal:
@Fryguy I realised rubocop wasn't updated in this pr, but `haml_lint` was i've put that back to the current version.

spec/rails_helper.rb
```diff
@@ -1,12 +1,14 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 ENV["RAILS_ENV"] ||= 'test'
 require File.expand_path("../../config/environment", __FILE__)
+require 'rails_helper'
```
Fryguy: it looks like this is requiring rails_helper, but the file was also renamed, so it's requiring itself?

spec/lib/bugzilla_service_spec.rb
```diff
@@ -1,5 +1,5 @@
 # rubocop:disable Style/NumericLiterals, Layout/IndentArray
-require 'spec_helper'
+require 'rails_helper'
```
Fryguy: I think let's go back to spec_helper, but in addition, just drop the require 'spec_helper' from every file because it's not needed if spec_helper is set in the .rspec file.

juliancheal:
@Fryguy @bdunne All green, everything looking good?

NickLaMuro:
@juliancheal I was working on a few things for the bot, and saw my old issue was still open:

https://github.com/ManageIQ/miq_bot/issues/427

So I created a PR based on your commits to get this effort up to snuff.  You should still be credited for the commits in that PR:

https://github.com/ManageIQ/miq_bot/pull/466

Hopefully that is okay!  I think this should be able to be closed now.

PR #448, opened 2019-10-14
Add bundle_inject

Fryguy:
@bdunne Please review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,8 @@
 source 'https://rubygems.org'
 
+plugin 'bundler-inject'
+require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
+
 gem 'rails', '~> 6.1.0', '>= 6.1.7.6'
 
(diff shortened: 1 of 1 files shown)


Fryguy:
I can't understand this one.

Gemfile
```diff
@@ -1,5 +1,8 @@
 source 'https://rubygems.org'
 
+plugin 'bundler-inject'
```
kbrock, line 3: did you want to version this? I think most of our spots we have "~> 2.0"

kbrock (approved):
just that question - otherwise, LGTM

Fryguy:
@kbrock It doesn't make its way to Gemfile.lock, so technically it doesn't matter.  I still can't understand why this fails though.

kbrock:
> @kbrock It doesn't make its way to Gemfile.lock, so technically it doesn't matter. I still can't understand why this fails though.

@Fryguy Is it possible that one of the dependencies mentions this at a different version?

kbrock:
kicking

PR #447, opened 2019-10-02
[super wip don't even look at it] Add command to test repos with the cross-repo tool

d-m-u:
Adding a bot action to test with the cross-repo magic.

It's a README change, don't look at it. I was just excited because it's freaking sweet. 

https://github.com/ManageIQ/miq_bot/issues/446

Issue #446, opened 2019-10-02
Cross Repo Testing


Fryguy:
The bot should accept a command on a PR to allow for cross-repo testing.

Expected message format:

`@miq-bot run-tests [<repos-to-test>] [including <gem-repos>]`

where:
- a repo is of the form `[org/]repo[@ref|#pr]`
- `repos-to-test` is a comma separated list of repos
- `gem-repos` is a set of gem repos

---

When the bot sees this command it will create a commit/PR against the cross-repo app [1].  The commit contents will modify the .travis.yml and set the env vars to match what was requested in the command

[1] presently in https://github.com/Fryguy/manageiq-cross_repo, but expected to be moved soon (and possibly split into separate library and execution repos).

---

Examples

- In ManageIQ/manageiq PR `#1234`:

  `@miq-bot run-tests manageiq-ui-classic`

  will create a commit with the .travis.yml changes:

  ```yaml
  env:
    global:
    - REPOS=ManageIQ/manageiq#1234,Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature
    matrix:
    - TEST_REPO=manageiq-ui-classic
  ```

- In ManageIQ/manageiq PR `#1234`:

  `@miq-bot run-tests manageiq-ui-classic,manageiq-api including Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature`

  will create a commit with the .travis.yml changes:

  ```yaml
  env:
    global:
    - REPOS=ManageIQ/manageiq#1234,Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature
    matrix:
    - TEST_REPO=manageiq-ui-classic
    - TEST_REPO=manageiq-api
  ```

- In ManageIQ/manageiq-ui-classic PR `#1234`:

  `@miq-bot run-tests manageiq`

  will create a commit with the .travis.yml changes:

  ```yaml
  env:
    global:
    - REPOS=ManageIQ/manageiq-ui-classic#1234
    matrix:
    - TEST_REPO=ManageIQ/manageiq
  ```

---

Other:
- The command should only works on PRs, and fail on issues
- The command should only be executable by the PR author or any member of the ManageIQ org.

Fryguy:
@agrare Can you eyeball this and perhaps add more examples or fix the ones I have here?

agrare:
Some of the other cases I see are:

[...]

PR #454, opened 2019-12-06
Add cross_repo_test bot command

NickLaMuro:
Addresses https://github.com/ManageIQ/miq_bot/issues/446

Still needs some cleanup, but should handle

- Cloning a configured `*-cross_repo-tests` repo
- Parses args from the command using variations outlined in https://github.com/ManageIQ/miq_bot/issues/446
- Creating a new commit with the `.travis.yml` changes
- Pushes commit and opens a pull request

Uses ideas from the following PRs/branches (thanks to @d-m-u and @Fryguy ):

- https://github.com/ManageIQ/miq_bot/pull/446
- https://github.com/ManageIQ/miq_bot/pull/447
- https://github.com/Fryguy/miq_bot/commit/dc0cc52ca6bbbfeaf5061c6681f4e6c53b9e85ab#diff-b351efdefa95964e43ea7fcabca89843

Notes
-----

- Decent amount of test helpers added to facilitate testing this as much as possible
- Under the assumption we won't be "monitoring" this repo like the others (doesn't really make sense)
- Found that most of the existing "`git` stuff" didn't really fit our needs

Regarding the last bit, I decided for this first pass, instead of trying to make work what was there for my needs, I would get a working prototype in place using `Rugged` directly.  With the heavy handed tests in place, swapping thigns out in the future should be fine, and the underlying implementation should still work as expected.

QA Steps
--------

~~Will add proper steps for this later, and I still haven't "done it live" myself (what I will be doing next), but the config changes that will be necessary are:~~

Okay, I most likely am going to miss some steps here since it is a bit of an involved setup for all of this, but here is a basic setup:
  
1. Create an organization on github that you own.  Example:  https://github.com/GrumpyTech/
  
2. Create a bot account.  Example: https://github.com/grumpynick-bot
  
3. Fork the `manageiq-cross_repo-tests` repository.

4. Ensure the bot account is a collaborator with `write` permissions on this repo
  
5. Fork a few other repos in the `ManageIQ` org for testing:
     - https://github.com/GrumpyTech/manageiq
     - https://github.com/GrumpyTech/manageiq-ui-classic
  
6. Configure the following in `config/settings/development.local.yml`:
  
  ```yaml
  github_credentials:
    email: <EMAIL_CONFIGURED_IN_GITHUB_FOR_THE_BOT>
    username: grumpynick-bot
    password: <PASSWORD>
  github:
    run_tests_repo:
      url: "<POINT_TO_YOUR_CLONE_URL_FOR_TESTING_USING_HTTPS>"
      name: manageiq-cross_repo-tests
  ```
  
7. `bundle exec foreman start` (no... not [**that** foreman](https://github.com/theforeman/foreman)...)
  
8. Create a PR for testing these commands and have at it!  https://github.com/GrumpyTech/manageiq-ui-classic/pull/1

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -1,0 +1,135 @@
+module GithubService
+  module Commands
+    # = GithubService::Commands::CrossRepoTest
+    #
+    # Triggers a build given the configured test repo:
+    #
+    #   Settings.cross_repo_tests_repo.*
+    #
+    # Which doing so will:
+    #
+    #    - validate the command
+    #    - create a branch
+    #    - create a commit with the travis.yml changes
+    #    - push said branch to the origin
+    #    - create a pull request for that branch
+    #
+    # More info can be found here on how the whole process works:
+    #
+    #   https://github.com/ManageIQ/manageiq-cross_repo-tests
+    #
+    # == Command structure
+    #
+    # @miq-bot cross-repo-test [<repos-to-test>] [including <extra-repos>]
+    #
+    # where:
+    #
+    #   - a `repo` is of the form [org/]repo[@ref|#pr]
+    #   - `repos-to-test` is a list of repos to have tested
+    #   - `extra-repos` is a list of repos (gems) to override in the bundle
+    #
+    # each "lists of repos" should be comma delimited.
+    #
+    # == Example
+    #
+    # In ManageIQ/manageiq PR #1234
+    #
+    #   @miq-bot cross-repo-test manageiq-api,manageiq-ui-classic#5678 \
+    #     including Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature
+    #
+    # will create a commit with the .travis.yml changes:
+    #
+    #   @@ -14,6 +14,6 @@ matrix:
+    #      fast_finish: true
+    #    env:
+    #      global:
+    #   -  - REPOS=
+    #   +  - REPOS=Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature,ManageIQ/manageiq#1234,manageiq-ui-classic#5678
+    #      matrix:
+    #   -  - TEST_REPO=
+    #   +  - TEST_REPO=ManageIQ/manageiq-api
+    #   +  - TEST_REPO=ManageIQ/manageiq-ui-classic#5678
+    #
+    # TODO:  Handle the "self" case, where `manageiq` is also a TEST_REPO
+    #
+    # (maybe include a "self" helper as well?)
+    #
+    class CrossRepoTest < Base
+      # The user calling the command
+      attr_reader :issuer
+
+      # The (extra) repo(s) being targeted to be included in the test run
+      attr_reader :repos
+
+      # The repo(s) that will have the test suite run
+      attr_reader :test_repos
+
+      # The arguments for the `cross-repo-test` command being called
+      attr_reader :value
+
+      restrict_to :organization
+
+      private
+
+      def _execute(issuer:, value:)
+        @issuer = issuer
+        parse_value(value)
+        return unless valid?
+
+        run_tests
+      end
+
+      def parse_value(value)
+        @value = value
+
+        @test_repos, @repos = value.split(/\s+including\s+/)
+                                   .map { |repo_list| repo_list.split(",") }
+
+        # Add the identifier for the PR for this comment to @repos here
+        @repos ||= []
+        @repos  << "#{issue.repo_name}##{issue.number}"
+
+        @test_repos.map! { |repo_name| normalize_repo_name(repo_name.strip) }
+        @repos.map!      { |repo_name| normalize_repo_name(repo_name.strip) }
+
+        [@repos, @test_repos].each(&:uniq!)
+      end
+
+      def normalize_repo_name(repo)
+        repo.include?("/") ? repo : "#{issue.organization_name}/#{repo}"
+      end
+
+      def valid?
+        unless issue.pull_request?
+          issue.add_comment("@#{issuer} 'cross-repo-test(s)' command is only valid on pull requests, ignoring...")
+          return false
+        end
+
+        true
+      end
+
+      def run_tests
+        create_cross_repo_test_branch
+        commit_yaml_changes
+        push_push_commit_to_remote
+        create_cross_repo_test_pull_request
+      end
+
+      def create_cross_repo_test_branch
+        # TODO
+      end
+
+      def commit_yaml_changes
+        # TODO
+      end
+
+      def push_push_commit_to_remote
+        # TODO
+      end
+
+      def create_cross_repo_test_pull_request
+        # TODO
+      end
+    end
+  end
+end
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -55,6 +55,9 @@ module GithubService
     # (maybe include a "self" helper as well?)
     #
     class CrossRepoTest < Base
+      # Reference to the branch we are creating off of origin/master
+      attr_reader :branch_ref
+
       # The user calling the command
       attr_reader :issuer
 
@@ -64,10 +67,58 @@ module GithubService
       # The repo(s) that will have the test suite run
       attr_reader :test_repos
 
+      # The *-cross_repo-tests rugged instance
+      attr_reader :rugged_repo
+
       # The arguments for the `cross-repo-test` command being called
       attr_reader :value
 
       restrict_to :organization
+
+      def self.test_repo_url
+        Settings.cross_repo_tests.url
+      end
+
+      def self.test_repo_name
+        Settings.cross_repo_tests.name
+      end
+
+      def self.test_repo_clone_dir
+        @test_repo_clone_dir ||= begin
+                                   url_parts = test_repo_url.split("/")[-2, 2]
+                                   repo_org  = url_parts.first
+                                   repo_dir  = test_repo_name
+                                   ::Repo::BASE_PATH.join(repo_org, repo_dir).to_s
+                                 end
+      end
+
+      def self.bot_name
+        Settings.github_credentials.username
+      end
+
+      def self.bot_email
+        Settings.github_credentials.email || "no_bot_email@example.com"
+      end
+
+      # The new branch name for this particular run of the command (uniq)
+      def branch_name
+        @branch_name ||= begin
+                           uuid     = SecureRandom.uuid
+                           bot_name = self.class.bot_name
+                           issue_id = "#{issue.repo_name}-#{issue.number}"
+
+                           "#{uuid}-#{bot_name}-run-tests-#{issue_id}"
+                         end
+      end
+
+      def run_tests
+        ensure_test_repo_clone
+        create_cross_repo_test_branch
+        update_travis_yaml_content
+        commit_travis_yaml_changes
+        push_commit_to_remote
+        create_cross_repo_test_pull_request
+      end
 
       private
 
@@ -108,27 +159,119 @@ module GithubService
         true
       end
 
-      def run_tests
-        create_cross_repo_test_branch
-        commit_yaml_changes
-        push_push_commit_to_remote
-        create_cross_repo_test_pull_request
+      ##### run_tests steps #####
+
+      # Clone repo (if needed) and initialize @rugged_repo
+      def ensure_test_repo_clone
+        repo_path = self.class.test_repo_clone_dir
+        if Dir.exist?(repo_path)
+          @rugged_repo = Rugged::Repository.new(repo_path)
+        else
+          url = self.class.test_repo_url
+          @rugged_repo = Rugged::Repository.clone_at(url, repo_path, :bare => true)
+        end
+        git_fetch
       end
 
       def create_cross_repo_test_branch
-        # TODO
+        @branch_ref = rugged_repo.create_branch(branch_name, "origin/master")
       end
 
-      def commit_yaml_changes
-        # TODO
+      # A lot of this is borrowed from some excellent work by Madhu:
+      #
+      #   https://github.com/ManageIQ/manageiq/blob/06de0607/lib/git_worktree.rb#L102-L110
+      #
+      def update_travis_yaml_content
+        raw_yaml = rugged_repo.blob_at(branch_ref.target.oid, ".travis.yml").content
+        content  = YAML.safe_load(raw_yaml)
+
+        content["env"] = {} unless content["env"]
+        content["env"]["global"] = ["REPOS=#{repos.join(',')}"]
+        content["env"]["matrix"] = test_repos.map { |repo| "TEST_REPO=#{repo}" }
+
+        entry = {}
+        entry[:path]  = ".travis.yml"
+        entry[:oid]   = @rugged_repo.write(content.to_yaml, :blob)
+        entry[:mode]  = 0o100644
+        entry[:mtime] = Time.now.utc
+
+        rugged_index.add(entry)
+        # rugged_index.write (don't do this...)
       end
 
-      def push_push_commit_to_remote
-        # TODO
+      def commit_travis_yaml_changes
+        bot       = self.class.bot_name
+        author    = {:name => issuer, :email => user_email(issuer),   :time => Time.now.utc}
+        committer = {:name => bot,    :email => self.class.bot_email, :time => Time.now.utc}
+
+        Rugged::Commit.create(
+          rugged_repo,
+          :author     => author,
+          :committer  => committer,
+          :parents    => [branch_ref.target],
+          :tree       => rugged_index.write_tree(rugged_repo),
+          :update_ref => "refs/heads/#{branch_name}",
+          :message    => <<~COMMIT_MSG
+            Running tests for #{issuer}
+
+            From Pull Request:  #{issue.fq_repo_name}##{issue.number}
+          COMMIT_MSG
+        )
+      end
+
+      def push_commit_to_remote
+        push_options = {}
+
+        if Settings.github_credentials.username && Settings.github_credentials.password
+          rugged_creds = Rugged::Credentials::UserPassword.new(
+            :username => Settings.github_credentials.username,
+            :password => Settings.github_credentials.password
+          )
+          push_options[:credentials] = rugged_creds
+        end
+
+        remote = @rugged_repo.remotes['origin']
+        remote.push(["refs/heads/#{branch_name}"], push_options)
       end
 
       def create_cross_repo_test_pull_request
-        # TODO
+        fq_repo_name = "#{issue.organization_name}/#{File.basename(self.class.test_repo_url, '.*')}"
+        pr_desc      = <<~PULL_REQUEST_MSG
+          From Pull Request:  #{issue.fq_repo_name}##{issue.number}
+          For User:           @#{issuer}
+        PULL_REQUEST_MSG
+
+        GithubService.create_pull_request(fq_repo_name,
+                                          "master", branch_name,
+                                          "[BOT] Cross repo test run", pr_desc)
+      end
+
+      ##### Duplicate Git stuffs #####
+
+      # Code that probably should be refactored to be shared elsewhere, but for
+      # now just shoving it here to get a working prototype together.
+
+      # Mostly a dupulicate from Repo.git_fetch (app/models/repo.rb)
+      #
+      # Don't need the credentials stuff since we are assuming https for this repo
+      def git_fetch
+        rugged_repo.remotes.each { |remote| rugged_repo.fetch(remote.name) }
+      end
+
+      def user_email(username)
+        GithubService.user(username).try(:email) || "no-name@example.com"
+      end
+
+      # Create a new Rugged::Index based off of "refs/remote/origin/master"
+      #
+      # Based off of GitWorktree in ManageIQ/manageiq
+      #
+      #   https://github.com/ManageIQ/manageiq/blob/06de0607/lib/git_worktree.rb#L395-L404
+      #
+      def rugged_index
+        @rugged_index ||= Rugged::Index.new.tap do |index|
+          index.read_tree(branch_ref.target.tree)
+        end
       end
     end
   end
--- a/spec/support/git_repo_helper.rb
+++ b/spec/support/git_repo_helper.rb
@@ -1,0 +1,228 @@
+require 'rugged'
+
+module GitRepoHelper
+  # NOTE:  This ripped from the work I did in  `juliancheal/code-extractor`:
+  #
+  #     https://github.com/juliancheal/code-extractor/pull/11
+  #
+  # (from the test/test_helper.rb)
+  #
+  # Which that was a modified form of the fake_ansible_repo.rb spec helper from
+  # the ManageIQ project:
+  #
+  #     https://github.com/ManageIQ/manageiq/blob/f8e70535/spec/support/fake_ansible_repo.rb
+  #
+  # It is being repurposed here to allow generating test repos in the specs
+  # without needing to do live clones, allowing for the testing of git actions
+  # in the specs.
+  #
+  # = TmpRepo
+  #
+  # This helper uses Rugged to create a stub git project for testing against
+  # with a configurable file structure.  To generate a repo, you just needs to
+  # be given a repo_path and a file tree definition.
+  #
+  #     file_tree_definition = %w[
+  #       foo/one.txt
+  #       bar/baz/two.txt
+  #       qux/
+  #       README.md
+  #     ]
+  #     TmpRepo.generate "/path/to/my_repo", file_tree_definition
+  #
+  #
+  # == File Tree Definition
+  #
+  # The file tree definition (file_struct) is just passed in as a word array for
+  # each file/empty-dir entry for the repo.
+  #
+  # So for a single file repo with a `foo.txt` plain text file, the definition
+  # as an array would be:
+  #
+  #     file_struct = %w[
+  #       foo.txt
+  #     ]
+  #
+  # This will generate a repo with a single file called `foo.txt`.  For a more
+  # complex example:
+  #
+  #     file_struct = %w[
+  #       bin/foo
+  #       lib/foo.rb
+  #       lib/foo/version.rb
+  #       test/test_helper.rb
+  #       test/foo_test.rb
+  #       tmp/
+  #       LICENSE
+  #       README.md
+  #     ]
+  #
+  # NOTE:  directories only need to be defined on their own if they are intended
+  # to be empty, otherwise a defining files in them is enough.
+  #
+  # == DSL methods
+  #
+  # When calling `TmpRepo.generate`, you can also pass a block to add content
+  # to files, move/remove files and directories, and make/tag commits as
+  # needed.
+  #
+  #   repo_structure = %w[
+  #     foo/bar
+  #     baz
+  #   ]
+  #
+  #   TmpRepo.generate dir, repo_structure do
+  #     update_file "foo/bar", "Bar Content"
+  #     commit "add Bar content"
+  #     tag "v1.0"
+  #
+  #     add_file "qux", "QUX!!!"
+  #     commit
+  #     tag "v2.0"
+  #   end
+  #
+  class TmpRepo
+    attr_accessor :repo
+    attr_reader   :file_struct, :last_commit, :repo_path, :index
+
+    delegate :create_branch, :checkout, :to => :repo
+
+    def self.generate(repo_path, file_struct = [], &block)
+      repo = new(repo_path, file_struct)
+      repo.generate(&block)
+      repo
+    end
+
+    def initialize(repo_path, file_struct)
+      @commit_count = 0
+      @repo_path    = Pathname.new(repo_path)
+      @name         = @repo_path.basename
+      @file_struct  = file_struct
+      @last_commit  = nil
+    end
+
+    def generate(&block)
+      build_repo
+
+      git_init
+      git_commit_initial
+
+      execute(&block) if block_given?
+    end
+
+    def clone(url, &block)
+      @repo        = Rugged::Repository.clone_at(url, @repo_path.to_s)
+      @index       = repo.index
+      @last_commit = repo.last_commit
+
+      execute(&block) if block_given?
+    end
+
+    # Run DSL methods for given TestRepo instance
+    def execute(&block)
+      instance_eval(&block)
+    end
+
+    def create_remote(name, dir)
+      Rugged::Repository.init_at dir, true      # create bare repo dir
+      repo.remotes.create(name, dir)            # add bare repo dir as remote
+      repo.remotes[name].push [repo.head.name]  # push current head to remote
+    end
+
+    # Commit with all changes added to the index
+    #
+    #   $ git add . && git commit -am "${msg}"
+    #
+    def commit(msg = nil)
+      git_add_all
+      @commit_count += 1
+
+      @last_commit = Rugged::Commit.create(
+        repo,
+        :message    => msg || "Commit ##{@commit_count}",
+        :parents    => [@last_commit].compact,
+        :tree       => index.write_tree(repo),
+        :update_ref => "HEAD"
+      )
+    end
+
+    def tag(tag_name)
+      repo.tags.create tag_name, @last_commit
+    end
+
+    # Add (or update) a file in the repo, and optionally write content to it
+    #
+    # The content is optional, but it will fully overwrite the content
+    # currently in the file.
+    #
+    def add_file(entry, content = nil)
+      path          = repo_path.join entry
+      dir, filename = path.split unless entry.end_with? "/"
+
+      FileUtils.mkdir_p dir.to_s == '.' ? repo_path : dir
+      FileUtils.touch path     if filename
+      File.write path, content if filename && content
+    end
+    alias update_file add_file
+
+    # Prepends content to an existing file
+    #
+    def add_to_file(entry, content)
+      path = repo_path.join entry
+      File.write path, content, :mode => "a"
+    end
+
+    def current_branch_name
+      repo.head.name.sub(/^refs\/heads\//, '')
+    end
+
+    private
+
+    # Generate repo structure based on @file_struct array
+    #
+    # By providing a directory location and an array of paths to generate,
+    # this will build a repository directory structure.  If a specific entry
+    # ends with a '/', then an empty directory will be generated.
+    #
+    # Example file structure array:
+    #
+    #     file_struct = %w[
+    #       foo/one.txt
+    #       bar/two.txt
+    #       baz/
+    #       qux.txt
+    #     ]
+    #
+    def build_repo
+      file_struct.each do |entry|
+        add_file entry
+      end
+    end
+
+    # Init new repo at local_repo
+    #
+    #   $ cd /tmp/clone_dir/test_repo && git init .
+    #
+    def git_init
+      @repo  = Rugged::Repository.init_at repo_path.to_s
+      @index = repo.index
+    end
+
+    # Add new files to index
+    #
+    #   $ git add .
+    #
+    def git_add_all
+      index.add_all
+      index.write
+    end
+
+    # Create initial commit
+    #
+    #   $ git commit -m "Initial Commit"
+    #
+    def git_commit_initial
+      commit "Initial Commit"
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 # Ignore all logfiles and tempfiles.
 /log/**/*.log
 /tmp
+/spec/tmp
 
 # Ignore binstubs
 /bin
--- a/README.md
+++ b/README.md
@@ -98,6 +98,15 @@ underscores replaced with hyphens.
 
   Example: `@miq-bot close_issue`
 
+- **`cross_repo_test test_repos* [including include_repos*]`**
+  Runs cross repo tests that include the current PR code base.
+
+  * This command is restricted to members of the organization containing the issue.
+  * Restricted use on pull requests.  This doesn't make sense to use on issues
+    since there is no code to run.
+
+  Example: `@miq-bot cross_repo_tests ManageIQ/manageiq#1234 including more_core_extensions@1234abcd`
+
 ## Development
 
 ### Prerequisites
--- a/spec/lib/github_service/commands/cross_repo_test_spec.rb
+++ b/spec/lib/github_service/commands/cross_repo_test_spec.rb
@@ -1,0 +1,175 @@
+require 'spec_helper'
+
+RSpec.describe GithubService::Commands::CrossRepoTest do
+  subject { described_class.new(issue) }
+
+  let(:issue)            { GithubService.issue(fq_repo_name, issue_id) }
+  let(:issue_id)         { 1234 }
+  let(:issue_url)        { "/repos/#{fq_repo_name}/issues/#{issue_id}" }
+  let(:issue_identifier) { "#{fq_repo_name}##{issue_id}" }
+  let(:fq_repo_name)     { "ManageIQ/bar" }
+  let(:command_issuer)   { "NickLaMuro" }
+  let(:command_value)    { "manageiq-ui-classic" }
+  let(:comment_url)      { "/repos/#{fq_repo_name}/issues/#{issue_id}/comments" }
+  let(:member_check_url) { "/orgs/ManageIQ/members/#{command_issuer}" }
+  let(:repo_check_url)   { "/orgs/ManageIQ/repos" }
+
+  before do
+    pr_fetch_response = single_pull_request_response(fq_repo_name, issue_id)
+    github_service_add_stub :url           => issue_url,
+                            :response_body => pr_fetch_response
+
+    github_service_add_stub :url             => member_check_url,
+                            :response_status => 204
+  end
+
+  describe "#execute!" do
+    def run_execute!(valid: true, add_stubs: true)
+      if add_stubs
+        # if we are stubbing, determine the number of expections based on validity
+        #
+        # Basically, if it isn't valid, we should never hit `run_tests`,
+        # otherwise it is run only once.
+        run_number = valid ? 1 : 0
+        expect(subject).to receive(:run_tests).exactly(run_number).times
+      end
+
+      subject.execute!(:issuer => command_issuer, :value => command_value)
+    end
+
+    it "runs tests when valid" do
+      run_execute!
+    end
+
+    context "with a non-member" do
+      let(:command_issuer)       { "non_member" }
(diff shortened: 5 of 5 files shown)


lib/github_service.rb
```diff
@@ -89,6 +89,20 @@ def milestones(fq_name)
...
+    # Cache repositories for an hour, and fetch them if a list they doesn't
+    # exist for a given user
+    def org_repos(org = nil, *args)
+      @org_repos_cache         ||= {}
+      @org_repos_cache_timeout ||= {}
+
+      if @org_repos_cache[org].nil? || @org_repos_cache_timeout[org] < Time.zone.now
+        @org_repos_cache_timeout[org] = 1.hour.from_now
+        @org_repos_cache[org]         = service.org_repos(org, *args).map(&:name)
+      end
```
Fryguy: Since the bot runs multi-threaded do we need a mutex to ensure there aren't races?

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      def self.test_repo_clone_dir
+        @test_repo_clone_dir ||= begin
+                                   url_parts = test_repo_url.split("/")[-2, 2]
+                                   repo_org  = url_parts.first
+                                   repo_dir  = test_repo_name
+                                   File.join(::Repo::BASE_PATH, repo_org, repo_dir)
+                                 end
+      end
+
+      # Name of the bot
```
agrare: Going to go out on a limb and say that this method name is self documenting :) A few of these comments are probably not necessary

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+        # Add the identifier for the PR for this comment to @repos here
+        @repos ||= []
+        @repos  << "#{issue.repo_name}##{issue.number}"
+
+        @test_repos.map! { |repo_name| normalize_repo_name(repo_name) }
+        @repos.map!      { |repo_name| normalize_repo_name(repo_name) }
+      end
+
+      def normalize_repo_name(repo)
+        repo.include?("/") ? repo : "#{issue.organization_name}/#{repo}"
```
agrare: We handle this internally in [`CrossRepo::Repository`](https://github.com/ManageIQ/manageiq-cross_repo/blob/master/lib/manageiq/cross_repo/repository.rb), can we just pass through the value that the user gives us?

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+          GithubService.org_repos(issue.organization_name).include?(name_only)
+        end
+      end
+
+      ##### run_tests steps #####
+
+      # Clone repo (if needed) and initialize @rugged_repo
+      def clone_repo_if_necessary
+        repo_path = self.class.test_repo_clone_dir
+        if Dir.exist?(self.class.test_repo_clone_dir)
```
agrare: Super minor, but can't you use `repo_path` here?

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+          File.write('.travis.yml', content.to_yaml)
+        end
+      end
+
+      def commit_travis_yaml_changes
+        index = rugged_repo.index
+        index.add('.travis.yml')
+        index.write
+
+        # rubocop:disable: Rails/TimeZone
```
agrare: Could you use `Time.now.utc` instead of disabling the rubocop check?

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+    #
+    #   @@ -14,6 +14,6 @@ matrix:
+    #      fast_finish: true
+    #    env:
+    #      global:
+    #   -  - REPOS=
+    #   +  - REPOS=manageiq#5678
+    #      matrix:
+    #   -  - TEST_REPO=
+    #   +  - TEST_REPO=manageiq-ui-classic#1234
```
agrare: Could we point to the manageiq-cross_repo README for examples instead of duplicating them here?  For the purposes of developers reading the class just an explanation of the ENV vars that need to be set is probably sufficient

NickLaMuro: I actually disagree with you here, because there is an added feature when using the bot of including the PR it is called from is automatically included in the `REPOS` section of the diff.  I think it is worth being clear about that.

I also just wanted the use cases that I was having to use as reference from https://github.com/ManageIQ/miq_bot/issues/446 (and are also out of date) somewhere close when building this so I didn't have to flip back and forth when developing this.

lib/github_service/commands/run_test.rb
NickLaMuro: Well, I was going to say I was just being consistent, but I guess I was only consistent when defining the `attr_readers`.  I probably am find removing this.

lib/github_service/commands/run_test.rb
agrare: That's fair, a sentence stating that the current PR will be added to the TEST_REPO list would probably accomplish the same thing

Fryguy: Updated the #446 yml diffs due to changes in the manageiq-cross_repo format (in particular the merging of CORE_REPO and GEM_REPOS into just REPOS)

lib/github_service/commands/run_test.rb
NickLaMuro: Probably?

To be fair, I was just copying what was done by @Fryguy here in this case:

https://github.com/Fryguy/miq_bot/commit/dc0cc52ca6bbbfeaf5061c6681f4e6c53b9e85ab#diff-b351efdefa95964e43ea7fcabca89843R26

That said, I at least want to make sure this works as is from `sidekiq` before changing this, since this is only a problem of being "too thorough" at the moment.

lib/github_service/commands/run_test.rb
NickLaMuro: Derp... yes...

lib/github_service/commands/run_test.rb
Fryguy: > because there is an added feature when using the bot of including the PR it is called from is automatically included in the REPOS section of the diff. I think it is worth being clear about that.

I agree with documenting that peculiarity, but that wasn't my takeaway from reading these comments.  Like @agrare I also viewed it as a copy/paste from the cross_repo examples, and am also concerned about [Basmati Rice](https://twitter.com/nzkoz/status/538892801941848064) (much like the examples in #446 became Basmati Rice)

> I also just wanted the use cases that I was having to use as reference from #446 (and are also out of date) somewhere close when building this so I didn't have to flip back and forth when developing this.

If it's temporary because this is still WIP and it's convenient, then yeah, this is fine.

lib/github_service/commands/run_test.rb
NickLaMuro: Yeah, probably.  I was a little weirded out doing that with a `git commit` versus something in the DB, but looking back, using `.utc` probably shouldn't be a problem.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+      def commit_travis_yaml_changes
+        index = rugged_repo.index
+        index.add('.travis.yml')
+        index.write
+
+        # rubocop:disable: Rails/TimeZone
+        bot       = self.class.bot_name
+        author    = { :name => issuer, :email => "no-name@example.com", :time => Time.now }
```
NickLaMuro: So just a note:  `Rugged` requires the email here, so I was just filling in a dummy example here.

In reality, it would be cool if the email registered from the user was used for this, since it would allow documenting in the commit both the bot creating it, and the user that triggered it.

That said, it is just a "cute" feature, and not really a hard requirement.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+      # Name of the bot
+      def self.bot_name
+        Settings.github_credentials.username
+      end
+
+      # The new branch name for this particular run of the command (uniq)
+      def branch_name
+        @branch_name ||= "#{self.class.bot_name}-run-tests-#{issue.number}-#{SecureRandom.uuid}"
```
Fryguy: I wonder if this should include the repo name as well or just remove the issue number?  issue number is just a number... e.g. in `miq-bot-1-01234abcdef` the `1` doesn't give you enough context, and doesn't help with uniqueness (since the uuid does that), so it's not contributing anything helpful to the naming by itself.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+      def run_tests
+        clone_repo_if_necessary
```
Fryguy: Minor naming style, but elsewhere we use the term ensure... i.e. `ensure_clone`

lib/github_service.rb
NickLaMuro: Well, we don't do that for any of the other caches here, so not sure for that here...

That said, now that you mention it, the `RunTest#run_tests` method probably needs a `mutex` since it is doing a `git checkout` effectively, and to calls at the same time could be an issue.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+        run_tests
+      end
+
+      def parse_value(value)
+        @value = value
+
+        @test_repos, @repos = value.split(/\s+including\s+/)
+                                   .map { |repo_list| repo_list.split(",") }
+
+        # Add the identifier for the PR for this comment to @repos here
```
Fryguy: Something about this wording feels missing...perhaps:

```ruby
# Ensure the identifier for the PR itself is added to the list of repos
```

On that note, could this end up duplicating the repo if it's already in the list?

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+        end
+
+        true
+      end
+
+      REPO_ONLY_REGEXP = /(?:[^\/]+\/)?(?<REPO_NAME_ONLY>[^@#]+)/.freeze
+      def repos_valid?(repos)
+        repos.all? do |repo|
+          name_only = repo.match(REPO_ONLY_REGEXP)[:REPO_NAME_ONLY]
+          GithubService.org_repos(issue.organization_name).include?(name_only)
```
Fryguy: I think this is wrong.  We may choose to test against forks (so different org) or gems that live on other orgs.  An example is ansible/ansible_tower_client_ruby which we maintain outside of the ManageIQ org, and may have as an included gem in cross_repo-tests.

I'm debating with myself whether we need a restriction at all.  Or perhaps only restrict against `@test_repos`

lib/github_service/commands/run_test.rb
NickLaMuro: > Updated the #446 yml diffs due to changes in the manageiq-cross_repo format (in particular the merging of CORE_REPO and GEM_REPOS into just REPOS)

Just to make sure with this as well, @agrare 's comment [here](https://github.com/ManageIQ/miq_bot/issues/446#issuecomment-538412470), while also suffering from the same "out of date" problems as the description, are also covered by the examples in the description now, right?  I was looking at the description trying parse if there was any difference, but didn't see any.

> am also concerned about Basmati Rice ...
>
> If it's temporary because this is still WIP and it's convenient, then yeah, this is fine.

@Fryguy @agrare agreeing with the "Basmati Rice" concern, so maybe we just have one example here as a middle ground (maybe the most complicated one)?

lib/github_service/commands/run_test.rb
Fryguy: or perhaps create an instance method called test_repo_clone_dir (or repo_path)?  I think having 2 names for the same thinis throwing me off)

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      ##### run_tests steps #####
+
+      # Clone repo (if needed) and initialize @rugged_repo
+      def clone_repo_if_necessary
+        repo_path = self.class.test_repo_clone_dir
+        if Dir.exist?(self.class.test_repo_clone_dir)
+          @rugged_repo = Rugged::Repository.new(repo_path)
+        else
+          url = self.class.test_repo_url
+          @rugged_repo = Rugged::Repository.clone_at(url, repo_path)
```
Fryguy: While I like using clone_at (and we should definitely use this to replace `MiniGitService.clone`!), I think this may not be thread-safe.  That is, if 2 threads try to clone at the same time, the exists check is not atomic....I'm not sure Rugged is thread safe at clone time.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+        git_fetch
+      end
+
+      def create_cross_repo_test_branch
+        rugged_repo.create_branch(branch_name, "origin/master")
+        rugged_repo.checkout(branch_name)
+      end
+
+      def update_travis_yaml_content
+        Dir.chdir(@rugged_repo.workdir) do
```
Fryguy: :-1: on using actual checkouts, though I know you did say you were trying to get things working first, and then would refactor later.  Actual checkouts create awful thread-safety issues.  In the rugged wrapper code in GitService we do everything with blobs and trees directly, specifically because it is all thread-safe...that is 2 branches can be created and worked on simultaneously within the git graph, whereas checkouts will collide, thus forcing serialization like we do in the MinigitService.

lib/github_service/commands/run_test.rb
NickLaMuro: Again, this was me mostly tweaking what @d-m-u had in her PR.

That said, if we move the `SecureRandom.uuid` to the beginning of the name, does that help?  Also, I probably should have added it as `#{issue.repo_name}_#{issue.number}`...

Fryguy: oh the uuid bit is fine and I don't really have an opinion on the location within the string...I'm more question the issue number.  I think `#{issue.repo_name}_#{issue.number}` would work fine

lib/github_service/commands/run_test.rb
Fryguy: I just noticed this is not a bare clone and it priobably should be.  It helps discourage use of the working dir, which otherwise leads to thread safety collision issues.

lib/github_service/commands/run_test.rb
Fryguy: The bot actually has an email, so we can use that if we make it a setting

lib/github_service/commands/run_test.rb
NickLaMuro: > EDIT: I misread the implementation and I see it handles forks, but I don't think it handles other arbitrary gems.

Yeah, I admit that I was being a bit "clever" there by removing the organization name to check.

> I'm debating with myself whether we need a restriction at all. Or perhaps only restrict against `@test_repos`

I think I agree that we should probably only do this against `@test_repos`, but I think it makes sense that the bot will only accept something that is within our org, since we don't have the control over how to run tests outside of our org and we can't assume consistency with how `rake` is configured.

lib/github_service/commands/run_test.rb
NickLaMuro: Yeah, maybe we should use a `Set` here, huh.

I guess using a normalized `repo_name` does have a purpose @agrare :)

lib/github_service/commands/run_test.rb
Fryguy: Oh I see you have the bot email below...yeah, let's make the bot email a setting (it's actually a gmail address).

For the author bit here, yeah, it would be kind of cool if we could get the comment requester's email, but it might be weird that one can get "commit credit" for something they didn't actually author.  Not sure how I feel about that :)

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+
+            From Pull Request:  #{issue.fq_repo_name}##{issue.number}
+          COMMIT_MSG
+        )
+      end
+
+      def push_commit_to_remote
+        push_options = {}
+
+        if Settings.github_credentials.username && Settings.github_credentials.password
```
Fryguy: What happens if we don't have creds?  Will this blow up?

Fryguy: Maybe [valid?](https://github.com/ManageIQ/miq_bot/pull/454/files#diff-2e4738d8e7840e1f2ad4b4161e5ddf2cR182) should check for this instead?  Not sure the best place to check for creds.

lib/github_service/commands/run_test.rb
Fryguy: Hmmm... I'm not sure the bot should be the one to make that determination.  That is, if I create a repo in my personal fork, say a new provider plugin I eventually want to introduce into manageiq, and it *does* have rake setup appropriately, then I feel the bot should be allowed to run it.

lib/github_service/commands/run_test.rb
NickLaMuro: Nope, since the tests I am running don't have creds either :)

Basically, it is similar to the implementation of `Repo#git_fetch` you wrote:

https://github.com/ManageIQ/miq_bot/blob/bc86fdec523510fc9c98fe363c708139a936feda/app/models/repo.rb#L133-L144

And will only add the options if they exist.  This means for clones that don't have credentials associated with them (like the file path ones I am using in the specs), this works when no credentials exist.

Granted, it will blow up if we are not authenticated with `github` properly (my current issue I am working through with [my testing at the moment](https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-562637555)).  Was trying to avoid setting up my bot with a separate SSH key, but it might be unavoidable, and we might need to go that route here instead, and assume a `ssh` based clone instead.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+          rugged_creds = Rugged::Credentials::UserPassword.new(
+            :username => Settings.github_credentials.username,
+            :password => Settings.github_credentials.password
+          )
+          push_options[:credentials] = rugged_creds
+        end
+
+        remote = @rugged_repo.remotes['origin']
+        remote.push(["refs/heads/#{branch_name}"], push_options)
```
Fryguy: We may have to be careful if the branch_name has a `/` in it...particularly if we insert the org/repo_name into the branch name as per other comments on the branch name structure.

lib/github_service/commands/run_test.rb
NickLaMuro: > though I know you did say you were trying to get things working first, and then would refactor later.

Agreed, and yes that was part of it.  If we can do that, then I will try to make that work then in a second pass.

The other bit I meant with that comment was I didn't want to try and re-purpose other "`git` stuffs" we already had in this repo just to get a POC working.  I think long term we probably should, and maybe before we merge this PR, but not for a first pass.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+      def create_cross_repo_test_pull_request
+        fq_repo_name = "#{issue.organization_name}/#{File.basename(self.class.test_repo_url, '.*')}"
```
Fryguy: Be aware that the test_repo_url may not actually be http based since it's just a remote and could be in any remote format...that is, in the miq-bot's deployment we'll likely use `git@github.com:ManageIQ/manageiq-cross_repo-tests` (which still works with your code, but I'm not sure you took that into account intentionally or not).

lib/github_service/commands/run_test.rb
NickLaMuro: Probably not, though, I specifically now remember that we were solving something like this for the `EmbeddedAnisble` work earlier this year, so I probably can just look at that for reference on what to do here 😄 .

lib/github_service/commands/run_test.rb
Fryguy: We may want to move this line to a method like `test_repo_name` or something and have it live up near test_repo_url

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      end
+
+      ##### Duplicate Git stuffs #####
+
+      # Code that probably should be refactored to be shared elsewhere, but for
+      # now just shoving it here to get a working prototype together.
+
+      # Mostly a dupulicate from Repo.git_fetch (app/models/repo.rb)
+      #
+      # Don't need the credentials stuff since we are assuming https for this repo
```
Fryguy: I think this is a wrong assumption.

lib/github_service/commands/run_test.rb
NickLaMuro: > oh the uuid bit is fine and I don't really have an opinion on the location within the string...

I was more suggesting that in it's current form, the `issue.number` and `uuid` get a bit muddled, and splitting them up to separate ends makes it a little more clear.

spec/support/git_repo_helper.rb
```diff
@@ -0,0 +1,274 @@
...
+  # Which that was a modified form of the fake_ansible_repo.rb spec helper from
+  # the ManageIQ project:
+  #
+  #     https://github.com/ManageIQ/manageiq/blob/f8e70535/spec/support/fake_ansible_repo.rb
+  #
+  # It is being repurposed here to allow generating test repos in the specs
+  # without needing to do live clones, allowing for the testing of git actions
+  # in the specs.
+  #
+  # = TmpRepo
```
Fryguy: Seems a bunch of stuff is still WIP in these comments?

lib/github_service/commands/run_test.rb
NickLaMuro: Yeah, that is a good point.  I will look into this.  Remind me if I forget after addressing some things, since this is important.

spec/support/git_repo_helper.rb
Fryguy: I wonder if we should make this entire class it's own gem.  I think an rspec-git_repo helper gem would be awesome.

spec/support/github_service_helper.rb
```diff
@@ -0,0 +1,217 @@
+require 'octokit'
+require 'github_service'
+require 'github_service/issue'
+
+module GithubServiceHelper
+  module_function
```
Fryguy: I have never seen this before.  Really interesting.  Not sure I like or not since lately I don't even like bare `private`/`protected` calls either.

lib/github_service/commands/run_test.rb
NickLaMuro: Well, it is similar to how they do email based merges, where someone sends an email of a patch, and the maintainer applies and merges it themselves.  In that case, the "author" is the person who wrote the patch (in this case, the person triggering the command), and the "committer" is the bot, who is actually doing the "merge".

lib/github_service/commands/run_test.rb
NickLaMuro: I just plan on inserting `issue.repo_name` for what it is worth, since `org` should be assumed since we restrict by that anyway.

spec/support/github_service_helper.rb
Fryguy: Oh weird...it creates the following methods as both static methods on the module *and also* private instance methods when mixed in.

spec/support/github_service_helper.rb
```diff
@@ -0,0 +1,217 @@
+require 'octokit'
+require 'github_service'
+require 'github_service/issue'
+
+module GithubServiceHelper
```
Fryguy: Out of curiosity, why use mocks when you can use VCR?

lib/github_service/commands/run_test.rb
NickLaMuro: > Be aware that the test_repo_url may not actually be http based...

As you found out here:

https://github.com/ManageIQ/miq_bot/pull/454/files#r354919027

I was sort of making that assumption in general for this particular git repo case.  Not sure if I agree with past self looking back on this, but for now, I would say assume that I expect that `manageiq-cross_repo-tests` has to be in it's https form for this current implementation.

> We may want to move this line to a method like `test_repo_name` or something and have it live up near test_repo_url

Well, since `self.test_repo_name` already exists, maybe `fq_test_repo_name`?  Probably has to be an instance method, though, I guess we could just configure it from the `url` in the class as well.  🤔

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,302 @@
...
+      attr_reader :repos
+
+      # The repo(s) that will have the test suite run
+      attr_reader :test_repos
+
+      # The *-cross_repo-tests rugged instance
+      attr_reader :rugged_repo
+
+      # The arguments for the `run-test` command being called
+      attr_reader :value
```
Fryguy: :issue and :value are technically on every command (and :org is also from the base? I'm not seeing it assigned here), so I think it's better to move those down to the base class.  Though I do understand that could introduce potential issues in the other commands, so I'd be fine with a separate PR for that (perhaps separated and then merged before this one, so this one can use it?)

spec/support/github_service_helper.rb
NickLaMuro: @Fryguy yeah, this is pure `Ruby` syntax, similar to how `FileUtils` works where you can either use the module directly, or include it in a class and they become private methods for internal use only.

lib/github_service/commands/run_test.rb
Fryguy: ah ok....yeah I didn't see that, but now that you mention it I agree.

lib/github_service/commands/run_test.rb
Fryguy: I just noticed we also have `restrict :organization` above, which prevents non-org members from asking the bot to do undesirable things, so if there was a concern that people could run random stuff via the bot, we are covered by that.

lib/github_service/commands/run_test.rb
Fryguy: *THAT'S* where I remember it from...I knew we were hitting this elsewhere and just couldn't remember :)

lib/github_service/commands/run_test.rb
Fryguy: As ok...makes sense :+1:

lib/github_service/commands/run_test.rb
Fryguy: Yeah for first pass this is fine, but we'll likely have to fix that on second pass.

Fryguy: `fq_test_repo_name` seems pretty good offhand

spec/support/github_service_helper.rb
NickLaMuro: I don't like VCR?

But in all seriousness, I do find VCR to be a heavy handed approach and try to avoid it when possible, since it creates a lot of bloat and churn in repos.  And it usually requires someone having specific access to an particular account to re-record when things change.

Regarding bloat and for some context, we would have a blob of this size in a yaml size somewhere just for fetching a single issue:

https://developer.github.com/v3/issues/#get-a-single-issue

Doing it this way, when things change (say, we update `octokit`), we just need to change a few lines if the expected JSON response has changed.  This is no better or worse (in my opinion) with what you would have to do with VCR.

* * *

To be fair, I don't think VCR is completely with out merit, I just try to avoid using it when possible.  I also find the `Faraday` mocking framework pretty nice and robust for my needs.

spec/support/github_service_helper.rb
NickLaMuro: @Fryguy yeah, this is pure `Ruby` syntax, similar to how `FileUtils` works where you can either use the module directly, or include it in a class and they become private methods for internal use only.

Fryguy:
@NickLaMuro Thanks for taking this on!  It's really coming together nicely.

lib/github_service/commands/run_test.rb
NickLaMuro: > :issue and :value are technically on every command

@Fryguy `:issue` is, but `:value` is passed in as a param to `_execute`, and is not an `attr_reader`.  I added that just so I didn't have to pass it around, though, I think `:value` as a `attr_reader` might be able to go away now that I think about it... 🤔 

> (and :org is also from the base? I'm not seeing it assigned here)

I think I was considering using `:org` at one point, and then I guess I just did `issue.organization` in some places instead and called it good.  I can just remove that `attr_reader` now probably.

spec/support/git_repo_helper.rb
NickLaMuro: > Seems a bunch of stuff is still WIP in these comments?

I think I mostly removed most of the parts we don't need, but I will do a once over.

> I wonder if we should make this entire class it's own gem. I think an rspec-git_repo helper gem would be awesome.

Agreed, though not sure if you want to wait on me writing another gem before we merge this :)

lib/github_service/commands/run_test.rb
NickLaMuro: > I just noticed we also have `restrict_to :organization`

@Fryguy that is actually from what _**YOU**_ wrote:

https://github.com/Fryguy/miq_bot/commit/dc0cc52ca6bbbfeaf5061c6681f4e6c53b9e85ab#diff-b351efdefa95964e43ea7fcabca89843R4

😆

NickLaMuro: > Hmmm... I'm not sure the bot should be the one to make that determination.

For what it is worth, this validation was an extension of what you originally had for some validations in your original spike commit:

https://github.com/Fryguy/miq_bot/commit/dc0cc52ca6bbbfeaf5061c6681f4e6c53b9e85ab#diff-b351efdefa95964e43ea7fcabca89843R14-R17

Not trying to suggest it is correct, just that is where I had originally pulled these from.

lib/github_service/commands/run_test.rb
NickLaMuro: Specifically with this one, I trimmed down the comment a bit to just one example, and tried to make it as all inclusive as possible.

NickLaMuro:
Addressed some of the comments, but still have a bit to go, but I am going to get something to eat...

Will work on this more later...

lib/github_service/commands/run_test.rb
NickLaMuro: Talked about above, so going to resolve this for now.  Undo if you disagree.

NickLaMuro:
Success!

https://github.com/GrumpyTech/manageiq-cross_repo-tests/pull/1

lib/github_service/commands/run_test.rb
NickLaMuro: > Granted, it will blow up if we are not authenticated with `github` properly (my current issue I am working through with [my testing at the moment](https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-562637555)). 

Okay, as mentioned here:

https://github.com/ManageIQ/miq_bot/pull/454#issuecomment-562762968

This was user (configuration) error, and works as expected when the bot has permissions write to that repo directly.

To avoid extra credentials (at least for dev), I think I am going to say this is good for now and resolve.  That said, I am fine with push back if you really want SSH key support instead, and I can look into adding that for this PR.  Just figured it is kind of redundant when we are already using user/pass auth with `octokit`.

lib/github_service/commands/run_test.rb
NickLaMuro: Okay, so I am trying to solve for this right now, but it has proven to be a pain as currently there isn't an easy way to parse the `user@blah:org/repo.git` ssh form while using something like `URI` (at least that I am aware of).

I will mess with this some more, but since I am out next week, I wouldn't be able to address something like this terribly well, so am leaning more towards "just use HTTPS" for now, and we can address the above in a followup.  But that is just my 2 cents.

lib/github_service/commands/run_test.rb
NickLaMuro: I think this is mostly address as is.

lib/github_service/commands/run_test.rb
NickLaMuro: I should have addressed this with the most recent commit.  🎉

lib/github_service/commands/run_test.rb
NickLaMuro: Done.  🎉

lib/github_service/commands/run_test.rb
NickLaMuro: Going to pass on working on this for this first pass, as it is being used in other ways to validate.

lib/github_service.rb
NickLaMuro: Okay, now that `git checkout` isn't happening, I think I have covered the bases that I was leaving this comment open for.  Resolving.

spec/support/git_repo_helper.rb
NickLaMuro: I think I cleaned up anything here that might have been "extra", so I think this is good.

Aside from me creating a whole new gem for this... but ain't nobody got time for dat...

lib/github_service.rb
```diff
@@ -89,6 +89,20 @@ def milestones(fq_name)
       milestones_cache[fq_name] ||= Hash[service.list_milestones(fq_name, :state => :all).map { |m| [m.title, m.number] }]
     end
 
+    # Cache repositories for an hour, and fetch them if a list they doesn't
```
d-m-u: sorry, this is a stupid nit but ```fetch them if a list they doesn't exist for a given user```

lib/github_service/commands/run_test.rb
Fryguy: Got it... my original PoC, was a PoC...I didn't put much thought in it, and mostly just ripped code from other places 😄

lib/github_service/commands/run_test.rb
Fryguy: Marking this one as unresolved for now just so we can track it that it's not been done.

Fryguy:
Master is now green, so can you rebase?

lib/github_service/commands/run_test.rb
```diff
@@ -177,33 +180,38 @@ def ensure_test_repo_clone
...
-          File.write('.travis.yml', content.to_yaml)
-        end
+        entry = {}
+        entry[:path]  = ".travis.yml"
+        entry[:oid]   = @rugged_repo.write(content.to_yaml, :blob)
+        entry[:mode]  = 0100644
+        entry[:mtime] = Time.now.utc
+
+        rugged_index.add(entry)
+        # rugged_index.write (don't do this...)
```
NickLaMuro: For the record:  This causes a segfault...

lib/github_service/commands/run_test.rb
NickLaMuro: @Fryguy and myself had a chat about this PR (and others), and I think on this issue, I suggested that we don't both with thread safety for this currently since it for our use case, we should only be cloning once.  After that, it will just be a `.fetch`, which should be thread safe.

Moving forward, when we refactor a decent amount of the git logic in this repo, I think making a clone method that is thread/process safe is a good idea, just not something that is needed for a single repo usage like we have here.

lib/github_service/commands/run_test.rb
NickLaMuro: So, just noting that we have made the decision to take this out.

That said (and I won't be trying to add this or a first pass), but I was thinking that it might be nice to be able to do:

```
@miq-bot run_tests manageiq-providers-*
```

As a shortcut, and having `org_repos` would allow for that, but we can solve that later, as it is more nice to have at this point.

lib/github_service/commands/run_test.rb
NickLaMuro: For this PR, we are going to skip this.

I will open up an issue after this PR is merged to resolve this.  It is doable, just don't think it is worth solving for at the moment and supporting both.

lib/github_service/commands/run_test.rb
NickLaMuro: As [mentioned above](https://github.com/ManageIQ/miq_bot/pull/454#discussion_r359090068), I will open up a issue once we merge this PR for handling this.  Once done, I will then resolve this again (basically, after merge though).

lib/github_service.rb
NickLaMuro: Fixed by removing 😉 

But yes... I suck at the English language...

NickLaMuro:
And re-ran a command with the rebased code, just to ensure it still worked:

https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-566805380
https://github.com/GrumpyTech/manageiq-cross_repo-tests/pull/6

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,278 @@
...
+      def self.test_repo_name
+        Settings.github.run_tests_repo.name
+      end
+
+      def self.test_repo_clone_dir
+        @test_repo_clone_dir ||= begin
+                                   url_parts = test_repo_url.split("/")[-2, 2]
+                                   repo_org  = url_parts.first
+                                   repo_dir  = test_repo_name
+                                   File.join(::Repo::BASE_PATH, repo_org, repo_dir)
```
Fryguy: There is now a class method `Repo.path` that gives you the path if you give it a qualified repo name (e.g. "SomeOrg/some_repo"), so that might be perfect here.

NickLaMuro: It will save about 5 characters...

```ruby
::Repo.path(File.join(repo_org, repo_dir))
```

Thought it was only going to be 1 characters, but it turns out that `Rugged::Repository` will take a `Pathname` instance, so `.to_s` wouldn't be required.

Though, what I will change it to is:

```ruby
::Repo::BASE_PATH.join repo_org, repo_dir
```

Since `BASE_PATH` is a `Pathname` and not a string.

NickLaMuro: Correction:  The above still needs to be a `String` since `Rugged::Repository.clone_at` _**does**_ require a string.

NickLaMuro:
@Fryguy Catching up on a few of my open PRs after the holidays.

Is there anything else you wanted me to do for this one?  I think it might be ready on my end, and I can always do a followup if there is something that needs a fix/refactor.

NickLaMuro:
cc @agrare regarding the above if you have any more input as well.

spec/lib/github_service/commands/run_test_spec.rb
```diff
@@ -0,0 +1,349 @@
...
+          For User:           @NickLaMuro
+        PR_BODY
+      }.to_json
+
+      github_service_add_stub :url           => "/repos/ManageIQ/cross_repo-tests-remote/pulls",
+                              :method        => :post,
+                              :request_body  => pull_request_data,
+                              :response_body => {"number" => 2345}.to_json
+
+      my_email = "NickLaMuro@loljustkidding.com"
```
bdunne: Please use `example.com` or one of the other domains reserved for documentation.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,278 @@
...
+    #    - push said branch to the origin
+    #    - create a pull request for that branch
+    #
+    # More info can be found here on how the whole process works:
+    #
+    #   https://github.com/ManageIQ/manageiq-cross_repo-tests
+    #
+    # == Command structure
+    #
+    # @miq-bot run-tests [<repos-to-test>] [including <extra-repos>]
```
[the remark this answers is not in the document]
NickLaMuro: I was going off the spec from @Fryguy in the issue ( https://github.com/ManageIQ/miq_bot/issues/446 ), so I will let him decided.

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,278 @@
...
+      # The *-cross_repo-tests rugged instance
+      attr_reader :rugged_repo
+
+      # The arguments for the `run-test` command being called
+      attr_reader :value
+
+      restrict_to :organization
+
+      def self.test_repo_url
+        Settings.github.run_tests_repo.url
```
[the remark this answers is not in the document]
NickLaMuro: Agreed.  Any suggestions on where you would like this?  Otherwise I can just put it as a top level setting.

lib/github_service/commands/run_test.rb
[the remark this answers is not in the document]
Fryguy: I have no preference on the actual command name.  run-tests was the name I used from forever ago, but cross-repo-test might be better now that we have an actual thing called cross-repo-test

lib/github_service/commands/run_test.rb
```diff
@@ -0,0 +1,278 @@
...
+    #     including Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature
+    #
+    # will create a commit with the .travis.yml changes:
+    #
+    #   @@ -14,6 +14,6 @@ matrix:
+    #      fast_finish: true
+    #    env:
+    #      global:
+    #   -  - REPOS=
+    #   +  - REPOS=Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature,ManageIQ/manageiq#1234
```
Fryguy, line 47: This should also probably have `manageiq-ui-classic#5678`, otherwise it's not an accurate test?

NickLaMuro: Oh yeah, yup.

I think this was a modified form of what you had in https://github.com/ManageIQ/miq_bot/issues/446 but I flubbed combining them together in a single example.  Will fix.

lib/github_service/commands/run_test.rb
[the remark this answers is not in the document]
NickLaMuro: Okay, if that is what we want, I can change it to `cross-repo-test` instead.

lib/github_service/commands/run_test.rb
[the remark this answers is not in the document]
NickLaMuro: This has been updated to be top level on `Settings`.  Resolving.

lib/github_service/commands/run_test.rb
NickLaMuro: Fixed.

lib/github_service/commands/run_test.rb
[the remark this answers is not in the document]
NickLaMuro: Okay, this has been updated, but going to re-seed my development database (blew it away by accident when working on a different PR) and try this again to make sure it still works.

NickLaMuro:
Doing a quick re-run of the "integration tests" that I did (similar to https://github.com/ManageIQ/miq_bot/pull/454#issuecomment-566806237 ) and will post that when I am done.  Aside from that, I think I have addressed the previous comments from above.

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -0,0 +1,278 @@
...
+      attr_reader :value
+
+      restrict_to :organization
+
+      def self.test_repo_url
+        Settings.cross_repo_tests.url
+      end
+
+      def self.test_repo_name
+        Settings.cross_repo_tests.name
```
NickLaMuro, line 83: This is the only change that I made after @Fryguy just approved, which was just:

```diff
- Settings.run_tests_repo.name
+ Settings.cross_repo_tests.name
```

agrare:
Hey @NickLaMuro @Fryguy @bdunne is this ready?

NickLaMuro:
@agrare On my end, yes.  I think I just made a change regarding this review:

https://github.com/ManageIQ/miq_bot/pull/454#discussion_r393994648

And wanted @Fryguy to double check that he was still good with it after the change.  I think this can be merged when people are satisfied.

PR #456, opened 2019-12-16
Fix Travis issues

Fryguy:
- Bump Travis Ruby version to 2.5.5
- Set up PostgreSQL properly

@bdunne Please review.


Fryguy:
This should get master back to green.

PR #469, opened 2020-01-14
[POC][WIP] Travis Build Failure monitor(ing)

NickLaMuro:
_Note: Currently still finalizing a lot here, so this is not ready for #PrimeTime™.  Just putting this up so folks know what is going on in the bot channel._

Adds a new worker that is in charge of monitoring Travis CI for build failures on monitored branches, and will post notifications in their respective channels.

This pull requests adds multiple classes/models/libs to help facilitate this functionality:

- The aforementioned "worker" (`TravisBranchMonitor`) class, that is mostly in charge of querying builds we care about, creating `BuildFailure` records as needed, and sending BuildFailureNotifications as needed.
- A `build_failures` table/model, which keeps track of previously failed builds (helps to avoid spam in chat, and we can post when a previously failing build has been fixed)
- `BuildFailureNotifier`:  A service object in charge of the sending messages to `gitter`.  Uses the `gitter-api-ruby` gem to do that.
- `TravisV3Client`:  A simple `Faraday` client lib that allows for using `v3` API endpoints with the `v2` API based `travis` gem.

Also updates the `travis` gem and adds `gitter-api-ruby`.


NickLaMuro:
**Update:** I think this is a pretty decent working copy for a first pass.

Currently this will:

- Monitor builds and post for failing builds on 'master' branches
- Only again post after it has been a day with no change
- Post that the build is passing once it does, and delete the `BuildFailure` record

Using the following config settings:

```yaml
---
# ...
gitter:
  api_url: http://localhost:5000
  api_prefix: "/api/v1"
# ...
travis_branch_monitor:
  included_repos:
    - ManageIQ/manageiq: not-real-group/community
    - GrumpyTech/manageiq-loggers: not-real-group/community
```

For the following builds for the [`GrumpyTech/manageiq-loggers` repo](https://github.com/GrumpyTech/manageiq-loggers):

- 💚 https://travis-ci.org/GrumpyTech/manageiq-loggers/builds/637174523
- 💚 https://travis-ci.org/GrumpyTech/manageiq-loggers/builds/637174582
- 🔴 https://travis-ci.org/GrumpyTech/manageiq-loggers/builds/637174979
- 💚 https://travis-ci.org/GrumpyTech/manageiq-loggers/builds/637189434

I got the following results running `bundle exec rake travis_branch_monitor:poll_single`:

Screen Shot 2020-01-14 at 7 17 23 PM

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,101 @@
...
+    # specific gitter room, so a hash is required in some cases.
+    #
+    # This override allows for doing this in the config
+    #
+    #   travis_branch_monitor:
+    #     included_repos:
+    #       - ManageIQ/manageiq-ui-classic: ManageIQ/ui
+    #       - ManageIQ/manageiq-gems-pending: ManageIQ/core
+    #       - ManageIQ/manageiq
+    #       - ManageIQ/miq_bot
```
Fryguy: Can this be simplified to

```yaml
travis_branch_monitor:
  included_repos:
    ManageIQ/manageiq-ui-classic: ManageIQ/ui
    ManageIQ/manageiq-gems-pending: ManageIQ/core
    ManageIQ/manageiq:
    ManageIQ/miq_bot:
```

This avoids the outer Array<Hash or String> and just makes it a Hash

db/migrate/20200109223351_create_build_failures.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBuildFailures < ActiveRecord::Migration
+  def change
+    create_table :build_failures do |t|
+      t.integer :repo_id
+      t.integer :branch_id
+      t.integer :travis_build_id
+
+      t.datetime :last_notified_at
+      t.timestamps
```
Fryguy: Could this just be a last_notified_at column on the branch and thus not need a whole table?

app/workers/travis_branch_monitor.rb
Fryguy: I think this simplifies a lof of the callers as well. For example, included_and_excluded_repos just becomes `super.keys`

db/migrate/20200109223351_create_build_failures.rb
NickLaMuro: Possibly.  I definitely took the approach of "haphazardly try and get this working" for this one, so most likely.  I also didn't want to just modify a table in case we wanted to separate concerns a bit.

That said, if the `travis_build_id` isn't being used here (definitely possible, since I forget if I even call it), then yes, I think we can get rid of this.  Or at least move those two columns to `Branch` and avoid the new model.  I will look into doing that today yet.

app/workers/travis_branch_monitor.rb
NickLaMuro: Agreed.  I think my thought process was that doing it as an array matches what was done with the other models more closely.  Not that it was a good idea to begin with.

I am all for making this simpler though, so I will muck with this and confirm that it works with the change just fine (most likely it will).

NickLaMuro: > For example, included_and_excluded_repos just becomes `super.keys`, I think :)

Not quite, since we always return an array of `[i, e]`, but still probably will be changing it.

The other thing that usually needs to be done is "stringify" the keys, since `Config` likes returning them as symbols, regardless of the yaml provided.

db/migrate/20200109223351_create_build_failures.rb
NickLaMuro: Moved the changes into the `Branch` model.  I did keep the `travis_build_id` (as `travis_build_failure_id`), but mostly worked just fine.

Thanks for the simplification suggestion!

NickLaMuro:
@Fryguy Addressed your changes.  Thanks for the feedback.

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,104 @@
+require 'travis'
+
+class TravisBranchMonitor
+  include Sidekiq::Worker
+  sidekiq_options :queue => :miq_bot_glacial, :retry => false
```
Fryguy: Should this be on glacial?  We generally only put the long running stuff on there, and this one seems pretty short.  @bdunne Thoughts?

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def process_branch(repo, branch)
+    branch_record = repo.regular_branches.where(:name => branch).first
+
+    # If we already have a failure record, call notify with that record
+    return branch_record.notify_of_failure if branch_record.previously_failing?
+
+    # otherwise, check if any builds exist with a failures, and if so, create a
+    # new BuildFailure record.
```
Fryguy: Some Basmati Rice in the comments here and above now that you don't have a separate table.

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def find_first_recent_failure(builds)
+    first_recent_failure = nil
+    builds.each do |build|
+      if build.passed?
+        break
+      else
+        first_recent_failure = build
+      end
+    end
```
Fryguy: I think you can do this a little cleaner with one of the following (assuming `builds` is Enumerable)

```ruby
builds.each_cons(2).detect { |_build, next_build| next_build.passed? }.first
# OR
builds.take_while(&:failed?).last
# OR
builds.slice_before(&:passed?).first.last
```

app/workers/travis_branch_monitor.rb
NickLaMuro: Totally was a "copy-pasta" here, so I have no opinions personally.

As far as being "quick", it depends on how many repos we are working with, but assuming the Travis API isn't haven't issues (which I have noticed while testing with just a single repo, FYI), it shouldn't take more than a minute to run this with the number of repos we probably would enable, but I didn't have a large sample size to test with.

app/workers/travis_branch_monitor.rb
NickLaMuro: Thanks.  👍 

Tried to catch some of this when I was refactoring yesterday, but clearly missed some spots.

app/workers/travis_branch_monitor.rb
NickLaMuro: The only reason I was doing it this way was to avoid multiple loop iterations (sorry, I come from the land of "performance"...), and was not able to recall some of the other iterator methods to do this in a simpler manner.

I would argue that the method name makes it clear what the intent is and since testing so far has been pretty manual for me so far, I am hesitant to make a change for something I know works currently.

That said, I will look into this, but if I push back (for now), that would be why.

NickLaMuro: Okay, took the second option since it responded the most like the original.  Added a test as well, though it is kind of crappy since it means to be able to remove the method all together (which kinda just makes sense at this point), we would need to remove the test.

Going to leave it for now so it is merge-worthy as is, but am looking at adding more tests in a way that would allow this refactoring to happen.  Can be done as a follow up though.

NickLaMuro:
> Use `update` instead of `update_attributes`

FYI:  These `rubocop` warnings are on lines that are working with `Travis::Client::Entity` models, and are **not** `ActiveRecord` models.

lib/build_failure_notifier.rb
```diff
@@ -0,0 +1,98 @@
...
+    @build     = repo.session.find_one(Travis::Client::Build, branch.travis_build_failure_id)
+  end
+
+  def post_failure
+    notification_msg = <<~MSG
+      > ### :red_circle: Build Failure in #{repo_branches_markdown_url}!
+      >
+      > **Travis Build**:  #{travis_build_url}
+    MSG
+    notification_msg << "> **Failure PR**:    #{offending_pr}\n" if offending_pr
```
bdunne: It's not always a PR to that repo which causes the failure. Sometimes a merge to core breaks the API repo, but it is only noticed after the API repo cron runs.

bdunne:
Also, does this need to be rebased due to the rails upgrade?

NickLaMuro:
> Also, does this need to be rebased due to the rails upgrade?

Shouldn't, as I just rebased it earlier today (fixing that exact problem)

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def process_repos
+    enabled_repos.each do |repo|
+      process_repo(repo)
+    end
+  end
+
+  def process_repo(repo)
+    # repo.regular_branch_names.each do |branch|
```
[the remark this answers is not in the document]
NickLaMuro: Most definitely.  This code definitely is subject to "copy-pasta syndrome".

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  include Sidetiq::Schedulable
+  recurrence { hourly.minute_of_hour(0, 15, 30, 45) }
+
+  include SidekiqWorkerMixin
+
+  class << self
+    private
+
+    # For this class, repos will sometimes the repo needs to be mapped to a
```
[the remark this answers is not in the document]
NickLaMuro: Okay... this one is probably just my brain being stupid... thanks.

app/models/branch.rb
```diff
@@ -99,4 +99,36 @@ def fq_branch_name
...
+
+  def should_notify_of_failure?
+    last_notified_at.nil? || last_notified_at < 1.day.ago
+  end
+
+  # If we have reported a failure before and the branch is now green.
+  #
+  # The other advantage of checking `last_notified_at.nil?` here is that we
+  # save a Travis API call, since we shouldn't be creating BuildFailure records
+  # without haven't found a build failure elsewhere (e.g. TravisBranchMonitor).
```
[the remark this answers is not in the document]
NickLaMuro: Probably...

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,95 @@
...
+  def process_repo(repo)
+    # repo.regular_branch_names.each do |branch|
+    # end
+
+    # TODO:  Handle this for all release branches
+    process_branch(repo, 'master')
+  end
+
+  def process_branch(repo, branch)
+    branch_record = repo.regular_branches.where(:name => branch).first
```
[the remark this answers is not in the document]
NickLaMuro: Maybe?  It has been a few months, and I might have just copied what was found elsewhere, but I will look into it.

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
NickLaMuro: Followup:  I didn't review this comment without looking at the context.

So I think this is related to the `TODO` that is below:

```ruby
# TODO:  Handle this for all release branches
process_branch(repo, 'master')
```

An probably what should be done is we should be doing `repo.regular_branches` (?) and passing the `branch_record` directly into `process_branch` instead of doing the lookup there.  That said, unsure if `regular_branches` does what we want in this case.  Would have to look.

That is also a question if we want to first start with 'master' for a first pass to test this feature out and then move on to the release branches going forward.  That way we can determine if this is indeed the route we want to take with this, or if it is causing too much noise in chat.

app/workers/travis_branch_monitor.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    # If we already have a failure record, call notify with that record
+    return branch_record.notify_of_failure if branch_record.previously_failing?
```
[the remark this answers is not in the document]
NickLaMuro: So this probably is a "naming is hard" problem, but `.notify_of_failure` actually does that.  It does the `.passing?` check which will determine if a repo is still failing for that given branch.

I do forget the rational for my logic as to "why" I split it up like this, so I will have to look at this again, but the logic does as you wish, it just is a little unclear.

lib/build_failure_notifier.rb
NickLaMuro: That is why the `if offending_pr` is there, which can return `nil`.

NickLaMuro: However, I am open to suggestions on how we can improve this to better provide context for the message.  I realize that this could still display a false positive by displaying the "last merged commit" on something that was merged weeks ago, so maybe only display the PR if it was merged recently?  Within the last 3 days?

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
NickLaMuro: See https://github.com/ManageIQ/miq_bot/pull/469/files#r393957986

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
bdunne: We probably don't want to track all of the `regular_branches` it probably only makes sense for a subset (master, jansa, ivanchuk, hammer) or something like that.  Maybe that belongs in a setting?

db/migrate/20200109223351_add_branch_build_failures.rb
```diff
@@ -0,0 +1,6 @@
+class AddBranchBuildFailures < ActiveRecord::Migration[5.2]
+  def change
+    add_column :branches, :travis_build_failure_id, :integer
+    add_column :branches, :last_notified_at, :datetime
```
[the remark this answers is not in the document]
NickLaMuro: I blame @Fryguy for the bad column name:

https://github.com/ManageIQ/miq_bot/pull/469#discussion_r367066112

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
NickLaMuro: Yeah, that probably makes sense.

This probably falls in the "doesn't need to be a private setting" category argument that I have been making, but we can adjust that at a later date.  I will look into making this a regular setting for now.  👍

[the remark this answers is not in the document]
Fryguy: > We probably don't want to track all of the regular_branches it probably only makes sense for a subset (master, jansa, ivanchuk, hammer) or something like that. Maybe that belongs in a setting?

We only track branches we care about, so I think all regular branches is ok.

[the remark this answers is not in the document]
bdunne: We might want to do some cleanup then because we are tracking everything back to darga.

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
NickLaMuro: Removed this line as it is now handled in `process_repo`

app/workers/travis_branch_monitor.rb
NickLaMuro: Done.

app/workers/travis_branch_monitor.rb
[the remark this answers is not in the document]
NickLaMuro: Currently using `.regular_branches` for this, so I think that has resolved this from a code aspect.  Leaving this unresolved just so it isn't lost as something we might want to look at on prod.

chessbyte:
@Fryguy now that we moved to GitHub Actions, assuming this should be closed?

PR #470, opened 2020-01-24
MiqBot AutoCorrect for AddLabel

NickLaMuro:
Adds `DidYouMean` support to `GithubService::Commands::AddLabel`, which will auto correct invalid labels, or at the very least provide options for incorrect labels.

This PR does slightly change the format of "incorrect label" issue comments.

Links
-----

- "Inspiration":  https://gitter.im/ManageIQ/manageiq/core?at=5e29e4591a1c2739e3e8b85f

### DIFF
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -64,8 +64,5 @@ module GithubService
   end
 end
 
-# Travis HACK v2
-begin
-  String.methosd
-rescue NoMethodError
-end
+# Travis HACK v2 (debugging)
+String.methosd
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -64,20 +64,17 @@ module GithubService
   end
 end
 
-# Travis HACK v2 (debugging)
-#
-#
-
-puts
-puts
-pp ENV
-puts
-puts
-pp RbConfig::CONFIG
-puts
-puts
-pp $LOAD_PATH
-puts
-puts
-
-String.methosd
+# HACK: Travis / `bundle install --path ...` compat
+begin
+  retry_require_dym = false
+  require 'did_you_mean'
+rescue LoadError => error
+  if retry_require_dym
+    raise error
+  else
+    $LOAD_PATH.push(*Dir[File.join(Gem.default_dir, "gems", "did_you_mean*", "lib")])
+    $LOAD_PATH.uniq!
+    retry_require_dym = true
+    retry
+  end
+end
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -5,6 +5,7 @@ module GithubService
 
       def _execute(issuer:, value:)
         valid, invalid = extract_label_names(value)
+        process_extracted_labels(valid, invalid)
 
         if invalid.any?
           message = "@#{issuer} Cannot apply the following label#{"s" if invalid.length > 1} because they are not recognized: "
@@ -29,6 +30,19 @@ module GithubService
 
         # Then see if any are *still* invalid and split the list
         label_names.partition { |l| GithubService.valid_label?(issue.fq_repo_name, l) }
+      end
+
+      def process_extracted_labels(valid_labels, invalid_labels)
+        labels = GithubService.labels(issue.fq_repo_name)
+        invalid_labels.reject! do |label|
+          corrections = DidYouMean::SpellChecker.new(:dictionary => labels).correct(label)
+
+          if corrections.count == 1
+            valid_labels << corrections.first
+          end
+        end
+
+        [valid_labels, invalid_labels]
       end
     end
   end
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -76,5 +76,8 @@ puts
 pp RbConfig::CONFIG
 puts
 puts
+pp $LOAD_PATH
+puts
+puts
 
 String.methosd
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/add_label.rb
```diff
@@ -1,20 +1,21 @@
...
-          message = "@#{issuer} Cannot apply the following label#{"s" if invalid.length > 1} because they are not recognized: "
-          message << invalid.join(", ")
-          issue.add_comment(message)
+          issue.add_comment(invalid_label_message(issuer, invalid))
         end
 
         if valid.any?
           valid.reject! { |l| issue.applied_label?(l) }
-          issue.add_labels(valid)
+          issue.add_labels(valid) unless valid.empty?
```
Fryguy: For readability, can we use `if valid.any?` as opposed to the double negative?

NickLaMuro: Sure, that works.  👍

Fryguy:
This is AWESOME.

NickLaMuro:
@Fryguy thanks!

Need to figure out how `DidYouMean` gets loaded, because I "did not mean" for tests to fail and they didn't for me locally, but `bundler` messes with the path.  Including it in the Gemfile isn't a good solution either since it Ruby loads it early, and so you can get a:

```console
$ bundle exec rake db:setup
...
You have already activated did_you_mean 1.2.0, but your Gemfile requires did_you_mean 1.3.0.  Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

error as well, even if someone is using `bundle exec`.

* * *

I got a bit frustrated looking into the "why" of this Saturday, but I will be looking at it more today.

Fryguy:
Since it comes with Ruby can we just *not* put it in the Gemfile?

lib/github_service/commands/add_label.rb
```diff
@@ -30,6 +31,37 @@ def validate_labels(label_names)
...
+        labels = GithubService.labels(issue.fq_repo_name)
+        invalid_labels.each do |bad_label|
+          corrections   = DidYouMean::SpellChecker.new(:dictionary => labels).correct(bad_label)
+          possibilities = corrections.map { |l| "`#{l}`" }.join(", ")
+
+          message << "* `#{bad_label}` "
+          message << "(Did you mean? #{possibilities})" if corrections.any?
+          message << "\n"
+        end
+        message << "\nAll labels for `#{issue.fq_repo_name}`:  https://github.com/#{issue.fq_repo_name}/labels"
```
Fryguy: Do you have a screenshot on what all this looks like?  I'm having trouble visualizing.

NickLaMuro: Yeah, I will try running a bit today locally, but I ran out of API calls when testing it last week... 

https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-578345729

I was starting and stopping `sidekiq` a bunch so that was breaking the cache everytime, so I used up my API limit very quickly.

I will run it again this morning and get you a screenshot.

NickLaMuro:
> Since it comes with Ruby can we just not put it in the Gemfile?

So I did that with the error that I linked, but I was using `did-you-mean-1.3.x` for my 2.6.3 version of ruby, but ruby 2.5 comes with `did-you-mean-1.2.x`.  It will be a last resort, but I don't want to include it in the Gemfile since it causes really confusing errors if you happen to be running on a different version of ruby.

NickLaMuro:
@Fryguy example of a working "auto correct", and failure message (need to create a "similar" label yet though to show multiple matches output):

https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-578347807

NickLaMuro:
@miq-bot add_label ivanchuk/

lib/github_service/commands/add_label.rb
NickLaMuro: Examples can now be found here:

https://github.com/GrumpyTech/manageiq-ui-classic/pull/1#issuecomment-578347807

NickLaMuro:
@Fryguy oh, this is interesting...

```console
$ cat did_you_mean_check_1.rb
require 'did_you_mean'
$ cat did_you_mean_check_2.rb
class Foo
  def bar
  end
end

Foo.new.ban
$ ruby did_you_mean_check_1.rb
$ bundle exec ruby  did_you_mean_check_1.rb
Traceback (most recent call last):
        1: from did_you_mean_check.rb:1:in `<main>'
did_you_mean_check.rb:1:in `require': cannot load such file -- did_you_mean (LoadError)
$ ruby did_you_mean_check_2.rb
Traceback (most recent call last):
did_you_mean_check.rb:8:in `<main>': undefined method `ban' for #<Foo:0x00007fa5c70b3b18> (NoMethodError)
Did you mean?  bar
$ bundle exec ruby did_you_mean_check_2.rb
Traceback (most recent call last):
did_you_mean_check.rb:8:in `<main>': undefined method `ban' for #<Foo:0x00007f8a63207660> (NoMethodError)
Did you mean?  bar
```

So yeah, looks like anything with `bundle exec ...` will not allow `did_you_mean` to load... but it somehow is loaded on error.

More investigation is needed...

NickLaMuro:
@Fryguy I *OBVIOUSLY* will rebase this prior to a merge, but I have found a solution.  While a bit hacky, avoids needing to add `did_you_mean` to the bundle.

Let me know what you think, but mentioned issue on `ruby/did_you_mean`  in the last commit (avoiding linking it here, but it is number 117) makes it pretty clear that there is some wonkiness when using `bundle install --path ...`, which Travis does do.

NickLaMuro:
❗️ 
😱

PR #473, opened 2020-02-26
Apply changes from the production box

bdunne:
- Upgrade rugged
- rescue failures trying to delete directories that don't exist
- switch back to fetching using minigit
- add new systemd scripts

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -147,16 +147,18 @@ class Repo < ActiveRecord::Base
 
   # TODO: Move this to GitService::Repo
   def git_fetch
-    require 'rugged'
-    rugged_repo = Rugged::Repository.new(path.to_s)
-    rugged_repo.remotes.each do |remote|
-      fetch_options = {}
+    # TODO: Figure out why this doesn't work on the bot using rugged
+    with_git_service { |git| git.fetch }
+    # require 'rugged'
+    # rugged_repo = Rugged::Repository.new(path.to_s)
+    # rugged_repo.remotes.each do |remote|
+    #   fetch_options = {}
 
-      username = extract_username_from_git_remote_url(remote.url)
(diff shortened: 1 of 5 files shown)


systemd/miq_bot-sidekiq.service
```diff
@@ -0,0 +1,18 @@
+Unit]
```
Fryguy: Should this have a leading `[`?

app/models/repo.rb
```diff
@@ -147,16 +147,18 @@ def synchronize_pr_branches(expected)
...
 
-      username = extract_username_from_git_remote_url(remote.url)
-      fetch_options[:credentials] = Rugged::Credentials::SshKeyFromAgent.new(:username => username) if username
+    #   username = extract_username_from_git_remote_url(remote.url)
+    #   fetch_options[:credentials] = Rugged::Credentials::SshKeyFromAgent.new(:username => username) if username
 
-      rugged_repo.fetch(remote.name, fetch_options)
-    end
+    #   rugged_repo.fetch(remote.name, fetch_options)
+    # end
```
Fryguy: Let's discuss before merge

systemd/miq_bot-sidekiq.service
bdunne: Yep, fixed :+1:

app/models/repo.rb
bdunne: Okay
- I wanted to push this into master rather than rebasing the commit on the box every time.
- I usually don't like to comment code and leave it there (since it's in the git history), but I wanted to make it very obvious that this needs to be fixed.
- Also, I was hoping to get this in soon so that I can more easily update the box to the latest master

Fryguy: Ah ok, so this is actually straight up broken and is the thing we have manually patched in the running bot.  Then I'm :+1:

PR #331, opened 2017-03-02
Stale issue/PR marker

chrisarcand:
Inspired by rails-bot (which in this case is nothing more than [a manual script](https://github.com/rafaelfranca/rails-tools/blob/master/bin/mark_stale_issues) run by Rafael every few months), this worker takes out the trash over the weekend and helps clean up old issues and pull requests in every enabled repo.

### Details

* A pull request or issue that hasn't been updated (commented on, whatever) in the last **6 months** is considered stale.
* Stale *issues* are marked with the 'stale' label and commented on.
* Stale *pull requests* are closed and commented on. 
* As with rails-bot, issues/pull requests can be 'pinned' - that is, if the label 'pinned' (or any other label that we deem worthy) is present on the issue, it will not be marked as stale or closed even if it's extremely old. This is useful for longstanding things that we know are a continuing issue but take a really long time to fix. 
* The StaleIssueMarker is run every Saturday, specifically, as it potentially uses a fair number of queries and not many people are working on a Saturday. Freshens up the repos for a lovely Monday morning 🌅 ☕️ 
* This worker is a little different in that I've split the concept of 'the worker that does the thing' and 'the worker that schedules the action to be taken with multiple workers'. That is, there's a `StaleIssueMarker` and a `Schedulers::StaleIssueMarker`. We should probably follow a pattern more like this with other things we do to do more things asynchronously. Thoughts?

*Why are you labeling issues but closing PRs?*  
* As assumed with rails-bot, issues should probably be at least glanced at before actually closed. If an issue is marked as 'stale', the finder of the issue should assume that unless they reproduce the issue soon, a committer can come along and just close it without having to spend the time to explain themselves too much. 
* Pull requests, on the other hand, have a true shelf life. If a PR isn't touched for months, there's a *very high* probability that it isn't even mergable and the contributor will probably just close it anyway - and even if it were to be eventually worked on again, it will probably be changed enough that the contributor should just make a brand new pull request anyway.

### Deployment

This includes tests, but even still, changing the state of hundreds of issues/PRs on the core repo is scary stuff 👻  Therefore, I'll run this on the miq_bot repo first to make sure it doesn't cause a chaos monkey 🐒

Issue #390, opened 2017-12-18
Stale issue checker should bump pinned issues


Fryguy:
Right now, pinned issues are essentially ignored forever.  However I think it would be useful if a pinned issue is stale for 6 months that miq-bot still comments on it, but with a different message.  Not sure on the message itself...maybe something like

> Bump :turtle:
> 
> This issue has not been updated for 6 months, but has been pinned, so it must be important.  This is a reminder to come back and take a look at it.

Or something like that 😄   cc @chrisarcand

mfeifer:
See https://github.com/orgs/ManageIQ/projects/12#card-30245102

mfeifer:
@NickLaMuro  here ya go!

NickLaMuro:
Asking for a bit of clarification here, but I think most of this makes sense:

> Stale and unmergeable should be closed

Not sure I agree with this.  Some times unmergable happens when another automated service updates something as simple as a gem:

https://github.com/ManageIQ/miq_bot/pull/468 => https://github.com/ManageIQ/miq_bot/pull/466#event-3058233724

So this seems like it might get a bit over-zealous for some folks.

> Nudge after 1 month of no activity

No problem with this.  Any specific copy we want to go with this?

Maybe just a simple "Bumb 🐢 " from the bot?  Possibly linking to a document about our traige process?

> Stale after 3 month of no activity
> After turns stale and 3 months, close the PR.

I assume this is meant to be an update from the current 6 months and year policies we currently have, right?

Fryguy:
> I assume this is meant to be an update from the current 6 months and year policies we currently have, right?

Yes, in the new triage calls we had to answer the question of how to handle stale PRs, so we decided that just closing PRs after 6 months of staleness wasn't right, hence the new rules.  If a PR was 6 months of no activity *and* unmergeable it was highly likely that the author abandoned it...there could be exceptions (like your PR) but the authors can still reopen.  In the case of your PR, we'd likely pin it, so it wouldn't happen anyway.

Fryguy:
> Any specific copy we want to go with this?

I would take the current copy for why the issue went stale and tweak it a little to say something along the lines of "this is a reminder that it's been #{time} since an update".  I like the idea of a link to the triage process.

NickLaMuro:
> bugs - if reproducable, add the verified label. If it goes stale, remove verified and stale.

[...]

PR #474, opened 2020-02-26
[StaleIssueMarker] Use GithubService.search_issues

NickLaMuro:
This is a pretty large refactoring of the StaleIssueMarker, but makes use of the `search_issues` api to reduce the number of API calls required to run this worker.

As a side effect, it also simplifies each of the methods in the class so they are a bit easier to digest.

Links
-----

* Github `/search/issues` endpoint docs: https://developer.github.com/v3/search/#search-issues-and-pull-requests
* Original implementation:  https://github.com/ManageIQ/miq_bot/pull/331

The original implementation PR includes some design context that I think is worth being aware of when reviewing.

### DIFF
--- a/app/workers/schedulers/stale_issue_marker.rb
+++ b/app/workers/schedulers/stale_issue_marker.rb
@@ -9,7 +9,7 @@ module Schedulers
     include SidekiqWorkerMixin
 
     def perform
-      enabled_repo_names.each { |r| ::StaleIssueMarker.perform_async(r) }
+      StaleIssueMarker.perform_async
     end
   end
 end
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -23,49 +23,50 @@ Feel free to reopen this pull request if these changes are still valid.
 Thank you for all your contributions!
   EOS
 
-  attr_reader :fq_repo_name
-
-  def perform(fq_repo_name)
-    @fq_repo_name = fq_repo_name
-    raise "The label #{STALE_LABEL_NAME} does not exist on #{fq_repo_name}" unless GithubService.valid_label?(fq_repo_name, STALE_LABEL_NAME)
-
-    GithubService.issues(fq_repo_name, :state => :open, :sort => :updated, :direction => :asc).each do |issue|
-      pinned = (issue.labels & PINNED_LABELS).present?
-
-      if issue.updated_at < stale_date && !pinned
-        if issue.pull_request?
-          closable_prs << issue
-        else
-          stale_issues << issue
-        end
+  def perform
+    GithubService.search_issues(*search_args).each do |issue|
+      if issue.pull_request?
+        close_pr(issue)
+      else
+        mark_as_stale(issue)
       end
-    end
-
-    stale_issues.each do |issue|
-      next if issue.labels.include?(STALE_LABEL_NAME)
-      logger.info("[#{Time.now.utc}] - Marking issue #{fq_repo_name}##{issue.number} as stale")
-      issue.add_labels([STALE_LABEL_NAME])
-      issue.add_comment(STALE_ISSUE_MESSAGE)
-    end
-
-    closable_prs.each do |pr|
-      logger.info("[#{Time.now.utc}] - Closing stale PR #{fq_repo_name}##{pr.number}")
-      GithubService.close_pull_request(pr.fq_repo_name, pr.number)
-      pr.add_comment(CLOSABLE_PR_MESSAGE)
     end
   end
 
   private
 
+  def search_args
+    query  = "is:open archived:false update:<#{stale_date.strftime('%Y-%m-%d')}"
+    query << " #{PINNED_LABELS.map { |label| %(-label:"#{label}") }.join(" ")}"
+    query << " #{enabled_repo_names.map { |repo| %(repo:"#{repo}") }.join(" ")}"
+
+    [query, {:sort => :updated, :direction => :asc}]
+  end
+
   def stale_date
     @stale_date ||= 6.months.ago
   end
 
-  def stale_issues
-    @stale_issues ||= []
+  def validate_repo_has_stale_label(repo)
+    unless GithubService.valid_label?(repo, STALE_LABEL_NAME)
+      raise "The label #{STALE_LABEL_NAME} does not exist on #{repo}"
+    end
   end
 
-  def closable_prs
-    @closable_prs ||= []
+  def mark_as_stale(issue)
+    return if issue.labels.include?(STALE_LABEL_NAME)
+
+    validate_repo_has_stale_label(issue.fq_repo_name)
+
+    logger.info("[#{Time.now.utc}] - Marking issue #{issue.fq_repo_name}##{issue.number} as stale")
+    issue.add_labels([STALE_LABEL_NAME])
+    issue.add_comment(STALE_ISSUE_MESSAGE)
+  end
+
+  def close_pr(pull_request)
+    validate_repo_has_stale_label(pull_request.fq_repo_name)
+    logger.info("[#{Time.now.utc}] - Closing stale PR #{pull_request.fq_repo_name}##{pull_request.number}")
+    GithubService.close_pull_request(pull_request.fq_repo_name, pull_request.number)
+    pull_request.add_comment(CLOSABLE_PR_MESSAGE)
   end
 end
--- a/lib/github_service.rb
+++ b/lib/github_service.rb
@@ -42,6 +42,10 @@ module GithubService
     end
     alias issues list_issues
 
+    def search_issues(*args)
+      service.search_issues(*args).items.map { |issue| Issue.new(issue) }
+    end
+
     def issue_comments(*args)
       service.issue_comments(*args).map do |comment|
         IssueComment.new(comment)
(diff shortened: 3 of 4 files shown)


app/workers/stale_issue_marker.rb
```diff
@@ -23,49 +23,49 @@ class StaleIssueMarker
...
+    [query, {:sort => :updated, :direction => :asc}]
+  end
+
   def stale_date
     @stale_date ||= 6.months.ago
   end
 
-  def stale_issues
-    @stale_issues ||= []
+  def repo_has_label(repo)
```
NickLaMuro: For what it is worth, I am not 100% sold on this, but decided as a start to try and keep the previous functionality as best as possible.

In my opinion, I think we should probably automate our label creation for repos so a check like this isn't required.

app/workers/schedulers/stale_issue_marker.rb
```diff
@@ -9,7 +9,7 @@ class StaleIssueMarker
     include SidekiqWorkerMixin
 
     def perform
-      enabled_repo_names.each { |r| ::StaleIssueMarker.perform_async(r) }
+      StaleIssueMarker.perform_async
```
NickLaMuro: As a result of making use of the `search_issues` endpoint, this does end up going against the previous implementation in https://github.com/ManageIQ/miq_bot/pull/331 .

However, I think that the reduced API calls and less dense `perform` end up making that argument less valid when using this approach.  By filtering by label and date as part of the search filter, we can eliminate entire calls if said repo doesn't even have open issues.

NickLaMuro:
@miq-bot add_reviewer @bdunne 
@miq-bot add_reviewer @Fryguy

app/workers/stale_issue_marker.rb
Fryguy: We do automate our label creation, but accidents happen...would prefer having the code be safe

app/workers/stale_issue_marker.rb
```diff
@@ -23,49 +23,50 @@ class StaleIssueMarker
...
+
+    logger.info("[#{Time.now.utc}] - Marking issue #{issue.fq_repo_name}##{issue.number} as stale")
+    issue.add_labels([STALE_LABEL_NAME])
+    issue.add_comment(STALE_ISSUE_MESSAGE)
   end
 
-  def closable_prs
-    @closable_prs ||= []
+  def close_pr(pull_request)
+    repo_has_label(pull_request.fq_repo_name)
```
Fryguy: The method name is weird to me...perhaps `validate_stale_label_presence`?

Fryguy:
Overall, I like it!  Seems like a much nicer approach.

app/workers/stale_issue_marker.rb
```diff
@@ -23,49 +23,50 @@ class StaleIssueMarker
...
-      logger.info("[#{Time.now.utc}] - Closing stale PR #{fq_repo_name}##{pr.number}")
-      GithubService.close_pull_request(pr.fq_repo_name, pr.number)
-      pr.add_comment(CLOSABLE_PR_MESSAGE)
-    end
   end
 
   private
 
+  def search_args
+    query  = "is:open archived:false update:<#{stale_date.strftime('%Y-%m-%d')}"
```
Fryguy: what  does `update:<date` entail?  Like, if someone pushes a commit, does that count? or perhaps adds a label or assignee?

app/workers/stale_issue_marker.rb
```diff
@@ -23,49 +23,50 @@ class StaleIssueMarker
...
+    query << " #{PINNED_LABELS.map { |label| %(-label:"#{label}") }.join(" ")}"
```
Fryguy: You can probably also remove already stale issues with `-label:stale`

app/workers/stale_issue_marker.rb
NickLaMuro: It is the query equivalent of:

`if issue.updated_at < stale_date && !pinned`

With the `!pinned` part happening below.

Regarding the dates specifically, more info can be found here:

https://help.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests#search-by-when-an-issue-or-pull-request-was-created-or-last-updated
https://help.github.com/en/github/searching-for-information-on-github/understanding-the-search-syntax#query-for-dates

For the example for `<` with dates is:

> Query | Example
> -- | --
> `<YYYY-MM-DD` | `cats pushed:<2012-07-05`  matches code with the word "cats" in repositories that were pushed to before July 5, 2012.
```

So translating to `update` should mean that anything with the most recent update happening before `6.mounts.ago`.

app/workers/stale_issue_marker.rb
NickLaMuro: I decided against doing that since this is adding `stale` as part of the query, since we will be closing some of them in the future, based on this:

https://github.com/ManageIQ/miq_bot/issues/390#issuecomment-577715811

In the end, the labels get included with the results, so there isn't an extra query that is being called by doing this, it is just trimming down the results a bit.

app/workers/stale_issue_marker.rb
NickLaMuro: For a future update, I might consider doing this in one fell swoop in `perform` just so we aren't have to check against the same repo every time (though, labels are cached, so it is more a "avoid extra useless work" fix instead of a "avoid `N+1`s"), but I think it is simpler for this pass to just keep it as is instead of rocking the boat with an extra fix.

Will resolve for now.

app/workers/stale_issue_marker.rb
NickLaMuro: Would `validate_repo_has_stale_label` work?  The `_presence` bit just feels weird to me.

#namingIsHard

NickLaMuro: I went with my suggestion because I am self centered.  #dealWithIt

(in all seriousness, if you don't approve, I can change to your suggestion, just went with something to keep things moving)

app/workers/stale_issue_marker.rb
Fryguy: I'm sorry...I worded my question poorly.  I'm more asking if checking "update" is sufficient for finding all of the things people can do with a PR.  For example if someone pushes a commit is "update" updated?  I'm concerned that we mark something stale even if someone does something to the PR because the search query is not aware of some specific kind of interaction.

NickLaMuro: @Fryguy nah, I don't think you worded it poorly, I was trying to answer both

>  I'm more asking if checking "update" is sufficient for finding all of the things people can do with a PR.

The [first link](https://help.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests#search-by-when-an-issue-or-pull-request-was-created-or-last-updated) above is the best I can do to answer that, where it says:

> You can filter issues based on times of creation, or when they were last updated. For issue creation, you can use the created qualifier; to find out when an issue was last updated, you'll want to use the updated qualifier.

But if that is as you described, I can only say "I think so?" at best... seems like it would be, but I agree with you in that I am slightly concerned if this is an equivalent.

* * *

I guess the best I could do is a test.  So starting with this:

[`org:ManageIQ label:stale updated:<2019-02-28 sort:updated-desc`](https://github.com/pulls?utf8=%E2%9C%93&q=org%3AManageIQ+label%3Astale+updated%3A%3C2019-02-28+sort%3Aupdated-desc)

I get this as the first result:  https://github.com/ManageIQ/manageiq-providers-redfish/issues/22

If I bump back the date to `2019-02-24`:

[`org:ManageIQ label:stale updated:<2019-02-24 sort:updated-desc`](https://github.com/pulls?utf8=✓&q=org%3AManageIQ+label%3Astale+updated%3A<2019-02-24+sort%3Aupdated-desc)

It is no longer found on the list.

That said, I don't see that as a good example, since we are dealing with an Issue that doesn't deal with commit events, so I decided to try it with `is:pr` just to limit it a bit more to "things with commits":

[`is:pr org:ManageIQ label:stale updated:<2019-02-28 sort:updated-desc`](https://github.com/pulls?utf8=✓&q=is%3Apr+org%3AManageIQ+label%3Astale+updated%3A<2019-02-28+sort%3Aupdated-desc)

And this was the PR that I got at the top:  https://github.com/ManageIQ/manageiq/pull/3870

(it is funny to see the `@oleg-bot` at work on that one 😄 )

That PR has a "updated on Nov 28, 2017" in the result, so if we updated the date to the `2017-11-29`:

[`is:pr org:ManageIQ label:stale updated:<2017-11-29 sort:updated-desc`](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+org%3AManageIQ+label%3Astale+updated%3A%3C2017-11-29+sort%3Aupdated-desc)

The PR is still in the results, but bumping it back a day:

[`is:pr org:ManageIQ label:stale updated:<2017-11-28 sort:updated-desc`](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+org%3AManageIQ+label%3Astale+updated%3A%3C2017-11-28+sort%3Aupdated-desc)

It is removed, and it looks like what was registered as an updated was a [branch delete action by GregB](https://github.com/ManageIQ/manageiq/pull/3870#event-1361997212), so I think we can be relatively sure that this will cover commits...

I think... check my logic on that break down...

Fryguy: This sounds good to me.  Thanks for the analysis.

app/workers/schedulers/stale_issue_marker.rb
Fryguy: Seems to make this entire class not necessary anymore, but we can keep for now.

PR #479, opened 2020-02-28
add_label/remove_label:  Prevent non-triage team assignment/removal

### LINKED ISSUE #475: [RFE] Botify request to backport
Process described [here](https://github.com/ManageIQ/guides/blob/master/backport_process.md).

We need an enhancement to miq-bot to convert requests to `add label release_to_backport_to/yes` to be interpreted as `add label release_to_backport_to/yes?` for anyone who is not part of the [triage team](https://github.com/orgs/ManageIQ/teams/core-triage).  If it is too hard to limit to that group, then do it for everyone and the people with the privileges to change labels without the bot will get things done that way.

Also, not sure if we should disallow removal of labels `release_to_backport_to/yes` and `release_to_backport_to/no`

NickLaMuro:
Starting with the Jansa release, we are changing the backport process to require code changes that are requesting a backport submit the `jansa/yes?` first to get approval from the triage team.  This means that setting the `jansa/yes` label (and similar labels from this release forward) will not be allowed.

For the purposes of the bot, this means that any labels that have been assigned a `*/yes` label (only `jansa/yes` for now), will be automatically converted to `jansa/yes?`.

Also, removing any of the `*/yes` and `*/no` labels is no longer possible via the bot, and the bot will respond with a warning if removing any of the `UNREMOVABLE` labels is attempted.

Links
-----

Fixes #475

### DIFF
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -1,6 +1,10 @@
 module GithubService
   module Commands
     class AddLabel < Base
+      UNASSIGNABLE = {
+        "jansa/yes" => "jansa/yes?"
+      }.freeze
+
       private
 
       def _execute(issuer:, value:)
@@ -32,6 +36,7 @@ module GithubService
 
       def process_extracted_labels(valid_labels, invalid_labels)
         correct_invalid_labels(valid_labels, invalid_labels)
+        handle_unassignable_labels(valid_labels)
 
         [valid_labels, invalid_labels]
       end
@@ -46,7 +51,12 @@ module GithubService
             valid_labels << corrections.first
           end
         end
+      end
 
+      def handle_unassignable_labels(valid_labels)
+        valid_labels.map! do |label|
+          UNASSIGNABLE.key?(label) ? UNASSIGNABLE[label] : label
+        end
       end
 
       def invalid_label_message(issuer, invalid_labels)
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -31,16 +31,22 @@ module GithubService
       end
 
       def process_extracted_labels(valid_labels, invalid_labels)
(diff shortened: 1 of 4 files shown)


lib/github_service/commands/add_label.rb
```diff
@@ -1,6 +1,10 @@
 module GithubService
   module Commands
     class AddLabel < Base
+      UNASSIGNABLE = {
+        "jansa/yes" => "jansa/yes?"
+      }
```
NickLaMuro: Note:  For now this is hard coded, but I think moving forward, we should maintain a list elsewhere for this.

That said, I didn't want to blanket block `*/yes` labels since this is intended for `jansa` and up only.

lib/github_service/commands/remove_label.rb
```diff
@@ -1,19 +1,31 @@
 module GithubService
   module Commands
     class RemoveLabel < Base
+      UNREMOVABLE = %w[jansa/yes jansa/no]
```
NickLaMuro: This one could probably be a a little more programmatic, but left it like this for now.

Side note: probably removing the ability to remove `*/backported` would probably be a good idea in general.

bdunne: > Side note: probably removing the ability to remove `*/backported` would probably be a good idea in general.

:+1: Sounds good to me.  What do you think @simaishi ?

simaishi: Agreed! The only time `/backported` label should get removed is when we revert the backport, in which case I manually set appropriate labels.

bdunne: We should probably disable adding the label too, right?  Anyone that can do the backporting should be able to manually add the label (not through bot commands)

NickLaMuro: Cool, I will follow some of this up in another PR.

👍

PR #480, opened 2020-02-28
Making Dan happy, one commit at a time...

### LINKED ISSUE #453: [RFE] Please add a "assign_reviewer" alias for "add_reviewer"
I continue to accidentally try to accidentally use the command "assign_reviewer" (instead of "add_reviewer") because that's the terminology that github currently uses.

Can we please add an alias so that either works?

NickLaMuro:
See https://github.com/ManageIQ/miq_bot/issues/453 and https://github.com/ManageIQ/miq_bot/issues/453#issuecomment-562178148

Fixes #453

### DIFF
--- a/lib/github_service/commands/add_reviewer.rb
+++ b/lib/github_service/commands/add_reviewer.rb
@@ -1,6 +1,8 @@
 module GithubService
   module Commands
     class AddReviewer < Base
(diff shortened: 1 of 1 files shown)


djberg96:
Thanks!

PR #481, opened 2020-03-03
Add CommitMetadataChecker

### LINKED ISSUE #414: Warn against user pings (@ + "username") in commit messages
GitHub notifications will spam a user everytime a commit shows up with a user's `@username` in the subject.

Having a simple check for each commit on push (like the `puts` check) can prevent this from happening, or at least give the review a heads-up when it happens.

NickLaMuro:
Adds `GithubService.username_lookup` and a batch worker (`CommitMessageUsernameChecker`) to look through commit messages for any potential usernames, and warns on valid matches.

Implementation Details
----------------------

This PR takes a few steps to avoid the concerns in the original issue ( #414 ), while still addressing the concerns from this comment:

> "People may want to put the @foo in code blocks though, which should be acceptable. Whatever detection we use should allow that."
> - https://github.com/ManageIQ/miq_bot/issues/414#issuecomment-375085251

1. The `GithubService.username_lookup` will attempt to avoid using an API call by first checking against the non-api user endpoint.  If a 404 is returned, we set the value for the username key in the `username_lookup_cache` to `nil`.  Subsequent checks will see this `nil` value and assume it is an "invalid username".  Causes a slower initial load for this method, but if a lot of different `@variables` are being used in a commit message, this is one way to avoid burning through our API cap quickly.

2. The second way to avoid extraneos API calls is the `USERNAME_REGEXP`, which attempts to follow the username requirements for github.  This means that `@foo_bar` will run through this processor without warning, but something like `@records` will most likely trigger a warning (yes... that is [a taken username](https://github.com/records))

That said `@foo` is most likely a valid user, among other typical "one word" variable names, so this will not catch everything.  That said, the offenses are treated as a `:low`, so this should be more of a "courteous reminder" than a hard requirement.

Fixes #414

### DIFF
--- a/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
@@ -1,0 +1,62 @@
+require 'rugged'
+
+module CommitMonitorHandlers::CommitRange
+  class GithubPrCommenter::CommitMetadataChecker
+    include Sidekiq::Worker
+    sidekiq_options :queue => :miq_bot_glacial
+
+    include BatchEntryWorkerMixin
+    include BranchWorkerMixin
+
+    def perform(batch_entry_id, branch_id, new_commits)
+      return unless find_batch_entry(batch_entry_id)
+      return skip_batch_entry unless find_branch(branch_id, :pr)
+
+      complete_batch_entry(:result => process_commits(new_commits))
+    end
+
+    private
+
+    def process_commits(new_commits)
+      @offenses = []
+
+      new_commits.each do |commit_sha, data|
+        check_for_usernames_in(commit_sha, data["message"])
+      end
+
+      @offenses
+    end
+
+    # From https://github.com/join
+    #
+    #     "Username may only contain alphanumeric characters or single hyphens,
+    #     and cannot begin or end with a hyphen."
+    #
+    # For the beginning and and, we do a positive lookbehind at the beginning
+    # to get the `@`, and a positive lookhead at the end to confirm their is
+    # either a period or a whitespace following the "var" (instance_variable)
+    #
+    # Since there can't be underscores in Github usernames, this makes it so we
+    # rule out partial matches of variables (@database_records having a
+    # username lookup of `database`), but still catch full variable names
+    # without underscores (`@foobarbaz`).
+    #
+    USERNAME_REGEXP = /
+      (?<=@)          # must start with a '@' (don't capture)
+      [a-zA-Z0-9]     # first character must be alphanumeric
+      [a-zA-Z0-9\-]*  # middle chars may be alphanumeric or hyphens
+      [a-zA-Z0-9]     # last character must be alphanumeric
+      (?=[\.\s])      # allow only variables without "_" (not captured)
+    /x.freeze
+
+    def check_for_usernames_in(commit, message)
+      message.scan(USERNAME_REGEXP).each do |potential_username|
+        next unless GithubService.username_lookup(potential_username)
+
+        group   = ::Branch.github_commit_uri(fq_repo_name, commit)
+        message = "Username `@#{potential_username}` detected in commit message. Consider removing."
+        @offenses << OffenseMessage::Entry.new(:low, message, group)
+      end
+    end
+  end
+end
--- a/lib/github_service.rb
+++ b/lib/github_service.rb
@@ -113,6 +113,21 @@ module GithubService
       assignees_cache.delete(fq_name)
     end
 
+    def username_lookup(username)
+      if username_lookup_cache.key?(username)
+        username_lookup_cache[username]
+      else
+        username_lookup_cache[username] ||= begin
+          case Net::HTTP.new("github.com", 443).tap { |h| h.use_ssl = true }.request_head("/#{username}")
+          when Net::HTTPNotFound then nil # invalid username
+          when Net::HTTPOK       then service.user(username)[:id]
+          else
+            raise "Error on GitHub with username lookup"
+          end
+        end
+      end
+    end
+
     private
 
     def service
@@ -149,6 +164,10 @@ module GithubService
 
     def assignees_cache
       @assignees_cache ||= {}
+    end
+
+    def username_lookup_cache
+      @username_lookup_cache ||= {}
     end
 
     def respond_to_missing?(method_name, include_private = false)
--- a/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
+++ b/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
@@ -1,0 +1,98 @@
+describe CommitMonitorHandlers::CommitRange::GithubPrCommenter::CommitMetadataChecker do
+  let(:batch_entry)        { BatchEntry.create!(:job => BatchJob.create!) }
+  let(:branch)             { create(:pr_branch) }
+  let(:commit_message_1)   { "Adds SUPER FEATURE.  BREAKS NOTHING!" }
+  let(:commit_message_2)   { "fix tests" }
+  let(:commit_message_3)   { "fix moar tests" }
+
+  let(:commits) do
+    {
+      "abcd123" => {"message" => commit_message_1, "files" => ["app/models/super.rb"]},
+      "abcd234" => {"message" => commit_message_2, "files" => ["spec/models/super_spec.rb"]},
+      "abcd345" => {"message" => commit_message_3, "files" => ["spec/models/normal_spec.rb"]}
+    }
+  end
+
+  let(:username_lookup_cache) do
+    {
+      "Fryguy"          => 123, # valid user
+      "NickLaMuro"      => 234, # valid user
+      "NickLaMura"      => nil, # mispelled invalid user
+      "Nick-LaMuro"     => nil, # invalid user with hyphen
+      "booksandauthors" => nil  # not a user (surprisingly)
+    }
+  end
+
+  before do
+    stub_sidekiq_logger
+    stub_job_completion
+
+    allow(GithubService).to receive(:username_lookup_cache).and_return(username_lookup_cache)
+  end
+
+  context "with basic commit messages" do
+    it "doesn't create any offenses" do
+      described_class.new.perform(batch_entry.id, branch.id, commits)
+
+      batch_entry.reload
+      expect(batch_entry.result.length).to eq(0)
+    end
+  end
+
+  context "with multiline commit messages" do
+    let(:commit_message_1) do
+      <<~COMMIT_MSG
+        In which the hero adds the greatest feature...
+
+        With this change, I made it so that by changing this:
+
+            @database_records = Books.all
+
+        Avoids a N+1 on authors by adding an .includes call
+
+            @booksandauthors = Books.all.includes(:author)
+
+        Saving us 100s of queries for our tiny bookstore app.
+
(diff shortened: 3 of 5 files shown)


NickLaMuro:
@miq-bot add_reviewer @Fryguy 
@miq-bot add_reviewer @jrafanie 

cc @bdunne @juliancheal @chrisarcand

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_message_username_checker.rb
```diff
@@ -0,0 +1,62 @@
+require 'rugged'
+
+module CommitMonitorHandlers::CommitRange
+  class GithubPrCommenter::CommitMessageUsernameChecker
```
bdunne: Should this be a more generic `CommitMetadataChecker`?  I could see wanting to check the commit message and other things too

NickLaMuro: Sure, I can change it to that.

NickLaMuro: Updated.

jrafanie:
The trophy for top :trollface: goes to @chrisarcand 

image

juliancheal:
LGTM 👍

chrisarcand:
LOL. Wow.

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
```diff
@@ -0,0 +1,63 @@
...
+    #
+    # Since there can't be underscores in Github usernames, this makes it so we
+    # rule out partial matches of variables (@database_records having a
+    # username lookup of `database`), but still catch full variable names
+    # without underscores (`@foobarbaz`).
+    #
+    USERNAME_REGEXP = /
+      (?<=@)          # must start with a '@' (don't capture)
+      [a-zA-Z0-9]     # first character must be alphanumeric
+      [a-zA-Z0-9\-]*  # alphanumeric have hyphens or slashes for middle chars
```
Fryguy: this comment reads funny to me... Perhaps: `middle chars may be alphanumeric, hyphens, or slashes`

lib/github_service.rb
```diff
@@ -113,6 +113,21 @@ def refresh_assignees(fq_name)
       assignees_cache.delete(fq_name)
     end
 
+    def username_lookup(username)
+      if username_lookup_cache.key?(username)
+        username_lookup_cache[username]
+      else
+        username_lookup_cache[username] ||= begin
+          case Net::HTTP.get_response(URI("https://github.com/#{username}"))
```
Fryguy: I feel like an octokit lookup would be cheaper here than a full http get of their profile.

Fryguy: Alternately a HEAD request would be cheap as well and would still give you response 200/404

lib/github_service.rb
```diff
@@ -113,6 +113,21 @@ def refresh_assignees(fq_name)
       assignees_cache.delete(fq_name)
     end
 
+    def username_lookup(username)
+      if username_lookup_cache.key?(username)
+        username_lookup_cache[username]
```
Fryguy: Why the double lookup?  That is the `||=` below handles this case, so the if/else is not needed.

spec/support/github_service_helper.rb
```diff
@@ -0,0 +1,207 @@
...
+  #        200,                              # response status
+  #        github_service_response_headers,  # response headers
+  #        {                                 # response body
+  #          "id" => 1,
+  #          # ...
+  #        }.to_json
+  #      ]
+  #   end
+  #
+  def github_service_stubs(&block)
```
Fryguy: Why are we adding these stubs?  I'd rather we didn't and for this case I don't see why we need it.

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+            @database_records = Books.all
+
+        Avoids a N+1 on authors by adding an .includes call
+
+            @booksandauthors = Books.all.includes(:author)
+
+        Saving us 100s of queries for our tiny bookstore app.
+
+        Thanks to @Fryguy and @Nick-LaMuro for this suggestion!
```
Fryguy: I love the prose, but it's distracting to the issue at hand...this whole thing can be 3-4 lines long and still test the feature.

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    it "returns one offense for each valid username" do
+      described_class.new.perform(batch_entry.id, branch.id, commits)
+
+      batch_entry.reload
+      expect(batch_entry.result.length).to eq(3)
+      expect(batch_entry.result.first).to have_attributes(
+        :group   => "abcd123",
+        :message => "Username `@Fryguy` detected in commit message. Consider removing."
```
Fryguy: Reading this, it might be helpful to give a reference to the commit in question...

    Username `@Fryguy` detected in commit message for commit a0d8f20d. Consider removing.

Perhaps even link the commit like we do in other error messages.

Fryguy:
> Causes a slower initial load for this method, but if a lot of different @variables are being used in a commit message, this is one way to avoid burning through our API cap quickly.

I can't imagine it happens all that often that it would even remotely affect the cap...especially since you are caching it.

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
NickLaMuro: So pretty sure that will happen with `group`, which is what I was doing here.

Instead of the filename here:

https://github.com/ManageIQ/miq_bot/pull/426#issuecomment-432861521

It would just print the commit SHA, which should auto link based on the autolink spec:

https://help.github.com/en/github/writing-on-github/autolinked-references-and-urls#commit-shas

That said, I might need to add it as `"#{issuer}@#{sha}"`... unsure.

lib/github_service.rb
NickLaMuro: No, it won't work with `nil`, because it will keep checking it.  `nil` in this case is used to denote that it is an invalid username.

lib/github_service.rb
NickLaMuro: Good idea, will change.

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
NickLaMuro: Just found out that this is a thing:

https://github.com/ManageIQ/miq_bot/blob/d5a8b5e00895c9fa449a42ca317801e4bbef8c07/app/models/branch.rb#L25-L27

So I will make use of that for the group, which should solve your concern.

NickLaMuro:
> I can't imagine it happens all that often that it would even remotely affect the cap...especially since you are caching it.

I mean... have you ever actually looked at my commit messages?

spec/support/github_service_helper.rb
NickLaMuro: FYI, this was merged in another PR.

Fryguy:
> I mean... have you ever actually looked at my commit messages?

Honestly I never read commit messages on PRs.  I always review PRs in full.

NickLaMuro:
> Honestly I never read commit messages on PRs. I always review PRs in full.

Well, kinda thought reading this was was a little important:

https://github.com/ManageIQ/miq_bot/pull/481/commits/044699881e12d99eac5d570f5c52fabd96835524

but maybe not...

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
```diff
@@ -0,0 +1,63 @@
...
+    private
+
+    def process_commits(new_commits)
+      @offenses = []
+
+      new_commits.each do |commit_sha, data|
+        check_for_usernames_in(commit_sha, data["message"])
+      end
+
+      @offenses
```
bdunne: This feels a bit awkward that `@offenses` is created and returned here, but `check_for_usernames_in` is the only one that adds to the array.  Would it be better to `new_commits.flat_map` instead?

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
```diff
@@ -0,0 +1,63 @@
...
+      (?<=@)          # must start with a '@' (don't capture)
+      [a-zA-Z0-9]     # first character must be alphanumeric
+      [a-zA-Z0-9\-]*  # middle chars may be alphanumeric or hyphens
+      [a-zA-Z0-9]     # last character must be alphanumeric
+      (?=[\.\s])      # allow only variables without "_" (not captured)
+    /x.freeze
+
+    def check_for_usernames_in(commit, message)
+      message.scan(USERNAME_REGEXP).each do |potential_username|
+        next if potential_username == "chrisarcand" # trolololololololo
```
bdunne: :-1:

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      )
+    end
+  end
+
+  context "with troll commit messages" do
+    let(:commit_message_3) do
+      <<~COMMIT_MSG
+        fixes moar tests
+
+        I forget how to rebase... #dealWithIt @chrisarcand
```
bdunne: Can this be one of the usernames from your valid but unregistered test cases from above?

bdunne: Nevermind, that test case is covered by `commit_message_1` above, please remove the whole commit.

NickLaMuro: and we have a winner!

(yes, that commit was a troll in disguise the entire time.  See commit message.)

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
NickLaMuro: @bdunne so, wanted to double check this, but this is pretty much the standard for the other two workers are doing things:

https://github.com/ManageIQ/miq_bot/blob/903eef83bb8404a7f7d8ff782ebb81c08ba68498/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_filename_checker.rb#L20-L26
https://github.com/ManageIQ/miq_bot/blob/903eef83bb8404a7f7d8ff782ebb81c08ba68498/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker.rb#L20-L26

So didn't really want break from convention for this one.

Do you mind if we address this in a follow up?

bdunne: > Do you mind if we address this in a follow up?

:+1: Fine by me

PR #482, opened 2020-03-03
[AddLabel/RemoveLabel] Allow triage team to add all labels

NickLaMuro:
Adds support for the triage team to add any labels.

Adds a concern `IsTeamMember` to handle shared code across `AddLabel` and `RemoveLabel`.

Triage team name is configurable in the settings.

Follow up from #479

**NOTE**:  I am re-using the `github_service_helper.rb` that I introduced in #454 (and also used in #481 as well), and I expect when one of those are merged, I will have to deal with the merge conflicts in the other branches.  It is an isolated commit, so rebasing it should be simple enough.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -21,6 +21,7 @@ module MiqBot
     # config.i18n.default_locale = :de
 
     config.eager_load_paths << Rails.root.join("app/workers/concerns")
+    config.eager_load_paths << Rails.root.join("lib/github_service/concerns")
     config.eager_load_paths << Rails.root.join("lib")
 
     console do
--- a/lib/github_service/concerns/is_team_member.rb
+++ b/lib/github_service/concerns/is_team_member.rb
@@ -1,0 +1,36 @@
+module IsTeamMember
+  def triage_member?(username)
+    IsTeamMember.triage_team_members.include?(username)
+  end
+
+  # List of usernames for the traige team
+  #
+  # Cache triage_team_members, and refresh cache every 24 hours
+  #
+  # Note:  This is created as a class method
+  #
+  cache_with_timeout(:triage_team_members, 24 * 60 * 60) do
+    if member_organization_name && triage_team_name
+      team = GithubService.org_teams(member_organization_name)
+                          .detect { |t| t.name == triage_team_name }
+
+      if team.nil?
+        []
+      else
+        GithubService.team_members(team.id).map(&:login)
+      end
+    else
+      []
+    end
+  end
+
+  module_function
+
+  def triage_team_name
+    @triage_team_name ||= Settings.triage_team_name || nil
+  end
+
+  def member_organization_name
+    @member_organization_name ||= Settings.member_organization_name || nil
+  end
+end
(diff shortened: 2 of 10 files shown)


config/application.rb
```diff
@@ -21,6 +21,7 @@ class Application < Rails::Application
     # config.i18n.default_locale = :de
 
     config.eager_load_paths << Rails.root.join("app/workers/concerns")
+    config.eager_load_paths << Rails.root.join("lib/github_service/concerns")
```
Fryguy: I thikn we can avoid this concerns directory by just putting it in `github_service/is_team_member.rb` directly, then just doing an `include GithubService::IsTeamMember`

NickLaMuro: We can, I was just trying to follow the `concerns/` convention.  I have hard opinions on where it lives.

lib/github_service/concerns/is_team_member.rb
```diff
@@ -0,0 +1,32 @@
...
+  extend ActiveSupport::Concern
+
+  def triage_member?(username)
+    self.class.triage_team_members.include?(username)
+  end
+
+  module ClassMethods
+    # List of usernames for the traige team
+    def triage_team_members
+      return @triage_team_members if defined? @triage_team_members
```
bdunne: We probably need to bust this cache occasionally to bring in anyone that was recently added to or removed from the team, right?

NickLaMuro: Probably, but this is where I don't have much insight into how the bot is run, since periodic worker restarts pretty much "could" handle that.  But I am not sure how often that happens.

We could also just bust the cache every 24 hours as well.

NickLaMuro:
@bdunne updated `more_core_extensions` to get us access to `cache_with_timeout` and updated the `IsTeamMember` module to make use of that.  It is no longer a concern since the defining of the singleton method made it weird to deal with in my head, so it is now just a shared cache between all classes that use this module, which is probably fine.

lib/github_service/concerns/is_team_member.rb
NickLaMuro: Resolved.  See https://github.com/ManageIQ/miq_bot/pull/482#issuecomment-597930848

spec/lib/github_service/commands/remove_label_spec.rb
```diff
@@ -55,19 +77,64 @@
...
+          expect(issue).to     receive(:remove_label).with("jansa/yes?")
+        end
       end
 
-      it "only removes the applied label" do
-        expect(issue).to     receive(:remove_label).with("wontfix")
-        expect(issue).not_to receive(:remove_label).with("jansa/no")
-        expect(issue).not_to receive(:remove_label).with("jansa/yes")
-        expect(issue).to     receive(:remove_label).with("jansa/yes?")
+      context "without a triage team" do
```
bdunne: Duplicate context name?

NickLaMuro: Yurp.  Firxed.

PR #483, opened 2020-03-05
[StaleIssueMarker] Better triage workflow

### LINKED ISSUE #390: Stale issue checker should bump pinned issues

### LINKED ISSUE #433: Modify the bot to auto-close issues with stale labels longer than xx months
The bot currently add the stale label, but these stale issues can linger around a lonnnnggg time.
After discussing with Jason and Brandon, we need to add auto-closure in this instance.

Here's some background: https://github.com/ManageIQ/miq_bot/pull/331

### LINKED ISSUE #437: Issue was closed after 6 months, but was not marked as stale beforehand
Hmmm, more bot weirdness:
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/234#event-2192793160

Shouldn't it have first been marked stale? Seems it closed the issue after 6 months with no updates.

NickLaMuro:
There are a lot of changes here, and the commits go into greater detail, but at a high level:

- Bumps marking as `stale` to 3 months
- Auto closes pull requests that are `stale` and `unmergeable`
- Handles Pull Requests and Issues in the same manner
- When a `verified` issue becomes `stale`, remove `verified` and mark as `requires triage`
- Misc refactorings and changes to support the above

Notes
-----

- The `requires triage` label was just a label name I picked, but can be changed to whatever
  - (when the label name is decided on, that will need to be set for all repos)
- Some code was moved around to be put in `GithubService::Issue` instead of in this class.  Mostly for readability, but I could see how it might be considered a controversial change.
- The copy for marking stale and closing stale could probably use some review.
- `RSpec` `doubles` became a bit of a pain to manage in the specs, so I switch them to `GithubService::Issue` objects with valid `Sawyer::Resource` objects under the hood.  There is a little more "under the covers" knowledge about `Octokit` that is required for this, but it avoids some over stubbing that would have had to be done as a result of some of the refactoring.  I could see this also being a bit of a controversial change.
- This probably could be broken into the subcomponents described previously ( https://github.com/ManageIQ/miq_bot/issues/390#issuecomment-577715811 ), but was easier to think about this problem all at once instead of trying to address in tiny pieces.  That said, if requested I can break this up into the individual RFE items with little effort, but will leave that up to the reviewers.

Also, there is still a very big question mark around the "nudging" concept that was requested as part of #390 that I am unsure how to tackle still.

### The "bump" feature problem

The main concern here is that once the bot introduces a comment, it actually updates the `updated_at` field, and changes the search criteria pretty considerably.

To illustrate, right now we make two searches:

- All non-pinned issues/PRs that are becoming `stale`, or have been stale now for 3 months
- Any PRs that are now `stale` and `unmergeable`

The first of the above actually takes care of two of our use cases, since they both share the 3 months time window for `updated_at`, and whenever a bot interacts with a issue/PR, it resets that timer.  As a result with it, we also don't have to filter based on the `stale` label in the query, and just handle what we do with the issues that we get back from the search.

If we implement the "nudging/bump" concept, we would then get a set of search queries that look something like this:

- Un-pinned issues/PRs that have been untouched for 1 month, less than 2 months, and aren't `stale` (handle nudging)
- Un-pinned issues/PRs that have been untouched for 2 months, and aren't `stale`    (handle newly stale)
- `stale` issues/PRs that have been untouched for 3 months`                      (handle old and stale)
- Any PRs that are now `stale` and `unmergeable`                                 (handle newly unmergeable stale PRs)

Doubles our number of queries, but more importantly, it makes the whole process a bit harder to wrap my head around, and I am sure it would be harder for others as well.  Feels like there is little gain to be had for a decent amount of a logic change, but am unsure what of #390 is considered more important at this point.

Will leave the decision of pursuing the "bump" feature further to reviewers/assignies.

Links
-----

- Fixes #390 (original RFE)
- Fixes #433
- Fixes #437

### DIFF
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -5,22 +5,14 @@ class StaleIssueMarker
   include SidekiqWorkerMixin
 
   # If an issue/pr has any of these labels, it will not be marked as stale or closed
-  PINNED_LABELS       = ['pinned'].freeze
-  SEARCH_SORTING      = {:sort => :updated, :direction => :asc}.freeze
-  STALE_ISSUE_MESSAGE = <<-EOS.freeze
-This issue has been automatically marked as stale because it has not been updated for at least 3 months.
+  PINNED_LABELS  = ['pinned'].freeze
+  SEARCH_SORTING = {:sort => :updated, :direction => :asc}.freeze
+  COMMENT_FOOTER = <<~FOOTER.sub("ManageIQ\n", "ManageIQ ").strip!
+    Thank you for all your contributions!  More information about the ManageIQ
+    triage process can be found in [the traige process documentation][1].
 
-If you can still reproduce this issue on the current release or on `master`, please reply with all of the information you have about it in order to keep the issue open.
-
-Thank you for all your contributions!
-  EOS
-  CLOSABLE_PR_MESSAGE = <<-EOS.freeze
-This pull request has been automatically closed because it has not been updated for at least 3 months.
-
-Feel free to reopen this pull request if these changes are still valid.
-
-Thank you for all your contributions!
-  EOS
+    [1]: https://www.manageiq.org/docs/guides/triage_process
+  FOOTER
 
   # Triage logic:
   #
@@ -94,15 +86,37 @@ Thank you for all your contributions!
   def comment_as_stale(issue)
     mark_as_stale(issue)
 
+    message = "This #{issue.type} has been automatically marked as stale " \
+              "because it has not been updated for at least 3 months.\n\n"
+    if issue.pull_request?
+      message << "If these changes are still valid, please remove the "    \
+                 "`stale` label, make any changes requested by reviewers " \
+                 "(if any), and ensure that this issue is being looked "   \
+                 "at by the assigned/reviewer(s)\n\n"
+    else
+      message << "If you can still reproduce this issue on the current " \
+                 "release or on `master`, please reply with all of the " \
+                 "information you have about it in order to keep the "   \
+                 "issue open.\n\n"
+    end
+    message << COMMENT_FOOTER
+
     logger.info("[#{Time.now.utc}] - Marking issue #{issue.fq_repo_name}##{issue.number} as stale")
-    issue.add_comment(STALE_ISSUE_MESSAGE)
+    issue.add_comment(message)
   end
 
   def comment_and_close(issue)
     mark_as_stale(issue)
 
+    message  = "This #{issue.type} has been automatically closed because it " \
+               "has not been updated for at least 3 months.\n\n"
+    message << "Feel free to reopen this #{issue.type} if "
+    message << "these changes are still valid.\n\n" if issue.pull_request?
+    message << "this issue is still valid.\n\n"     unless issue.pull_request?
+    message << COMMENT_FOOTER
+
     logger.info("[#{Time.now.utc}] - Closing stale PR #{issue.fq_repo_name}##{issue.number}")
     GithubService.close_pull_request(issue.fq_repo_name, issue.number)
-    issue.add_comment(CLOSABLE_PR_MESSAGE)
+    issue.add_comment(message)
   end
 end
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -41,10 +41,10 @@ Thank you for all your contributions!
     query << unpinned_query_filter
 
     GithubService.search_issues(query, SEARCH_SORTING).each do |issue|
-      if issue.stale?
+      if issue.stale? || issue.unmergeable?
         comment_and_close(issue)
       else
-        mark_as_stale(issue)
+        comment_as_stale(issue)
       end
     end
   end
@@ -88,14 +88,19 @@ Thank you for all your contributions!
     return if issue.stale?
 
     validate_repo_has_stale_label(issue.fq_repo_name)
+    issue.add_labels([stale_label])
+  end
+
+  def comment_as_stale(issue)
+    mark_as_stale(issue)
 
     logger.info("[#{Time.now.utc}] - Marking issue #{issue.fq_repo_name}##{issue.number} as stale")
-    issue.add_labels([stale_label])
     issue.add_comment(STALE_ISSUE_MESSAGE)
   end
 
   def comment_and_close(issue)
-    validate_repo_has_stale_label(issue.fq_repo_name)
+    mark_as_stale(issue)
+
     logger.info("[#{Time.now.utc}] - Closing stale PR #{issue.fq_repo_name}##{issue.number}")
     GithubService.close_pull_request(issue.fq_repo_name, issue.number)
     issue.add_comment(CLOSABLE_PR_MESSAGE)
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -9,7 +9,7 @@ class StaleIssueMarker
 
   STALE_LABEL_NAME = 'stale'.freeze
   STALE_ISSUE_MESSAGE = <<-EOS.freeze
-This issue has been automatically marked as stale because it has not been updated for at least 6 months.
+This issue has been automatically marked as stale because it has not been updated for at least 3 months.
 
 If you can still reproduce this issue on the current release or on `master`, please reply with all of the information you have about it in order to keep the issue open.
 
@@ -23,6 +23,10 @@ Feel free to reopen this pull request if these changes are still valid.
 Thank you for all your contributions!
   EOS
 
+  # Triage logic:
+  #
+  # - Stale after 3 month of no activity
+  #
   def perform
     GithubService.search_issues(*search_args).each do |issue|
       if issue.pull_request?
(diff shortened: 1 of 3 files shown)


NickLaMuro:
@miq-bot assign @mfeifer 
@miq-bot add_reviewer @Fryguy

Fryguy:
Why was `requires triage` added?  I don't think it's necessary.

> When an issue becomes stale and is verified, remove the [verified] label and mark it as [requires triage] to inform that triage team that this issue needs to be re-evaluated for relevancy.

Triage team will review all stale issues on a regular cadence anyway, so it would be preferable to keep the `verified` label as it was...in triage we will see it is `stale`+`verified` and make a decision on that information.

app/workers/stale_issue_marker.rb
```diff
@@ -5,22 +5,14 @@ class StaleIssueMarker
   include SidekiqWorkerMixin
 
   # If an issue/pr has any of these labels, it will not be marked as stale or closed
-  PINNED_LABELS       = ['pinned', GithubService::Issue::REQUIRES_TRIAGE].freeze
-  SEARCH_SORTING      = {:sort => :updated, :direction => :asc}.freeze
-  STALE_ISSUE_MESSAGE = <<-EOS.freeze
-This issue has been automatically marked as stale because it has not been updated for at least 3 months.
+  PINNED_LABELS  = ['pinned', GithubService::Issue::REQUIRES_TRIAGE].freeze
+  SEARCH_SORTING = {:sort => :updated, :direction => :asc}.freeze
+  COMMENT_FOOTER = <<~FOOTER.sub("ManageIQ\n", "ManageIQ ").strip!
```
Fryguy: I don't think this is needed ... Why not just leave the multi-line comment?

app/workers/stale_issue_marker.rb
```diff
@@ -5,22 +5,14 @@ class StaleIssueMarker
...
-Thank you for all your contributions!
-  EOS
-  CLOSABLE_PR_MESSAGE = <<-EOS.freeze
-This pull request has been automatically closed because it has not been updated for at least 3 months.
-
-Feel free to reopen this pull request if these changes are still valid.
-
-Thank you for all your contributions!
-  EOS
+    [1]: https://github.com/ManageIQ/guides/blob/master/triage_process.md
```
Fryguy: Prefer https://www.manageiq.org/docs/guides/triage_process

app/workers/stale_issue_marker.rb
```diff
@@ -5,68 +5,134 @@ class StaleIssueMarker
...
+    GithubService.close_pull_request(issue.fq_repo_name, issue.number)
+    issue.add_comment(message)
   end
 
-  def close_pr(pull_request)
-    validate_repo_has_stale_label(pull_request.fq_repo_name)
-    logger.info("[#{Time.now.utc}] - Closing stale PR #{pull_request.fq_repo_name}##{pull_request.number}")
-    GithubService.close_pull_request(pull_request.fq_repo_name, pull_request.number)
-    pull_request.add_comment(CLOSABLE_PR_MESSAGE)
+  def comment_copy_footer
```
Fryguy: empty method?

Fryguy:
Overall LGTM. aside from the removal of the "requires triage" label as mentioned above.

app/workers/stale_issue_marker.rb
NickLaMuro: Err... might have been me working out a few ideas, and then fudging it when I committed it (or a failed rebase).

Will look into it.

Fryguy:
> RSpec doubles became a bit of a pain to manage in the specs, so I switch them to GithubService::Issue objects with valid Sawyer::Resource objects under the hood. There is a little more "under the covers" knowledge about Octokit that is required for this, but it avoids some over stubbing that would have had to be done as a result of some of the refactoring. I could see this also being a bit of a controversial change.

Can you explain this a bit more...I saw in your other PRs you were going in this direction, and I don't understand why the need for something that appears more complicated at a glance.

Fryguy:
> Also, there is still a very big question mark around the "nudging" concept that was requested as part of #390 that I am unsure how to tackle still.

We can do that in a follow up I think.

NickLaMuro:
@Fryguy This is specifically what I was asking you for clarification on here:

https://github.com/ManageIQ/miq_bot/issues/390#issuecomment-591657256

Specifically around this part where I tried to interpret:

> "If a bug is verified, and reaches the stale time period, remove the verified label so that it can be re-triaged, but don't mark as stale."

Which, the `requires triage` was meant to be a signal that recently both were removed (per the requirements from that issue).  Also:

> in triage we will see it is stale+verified and make a decision on that information.

This is the opposite of what was asked of me to do here:

https://github.com/ManageIQ/miq_bot/issues/390#issuecomment-577715811


* * *

Please clarify the requirements before I try and argue this further, as I don't want to be coding something up where there isn't even an agreement on how this should function.

Fryguy:
Ah! thanks for the context...I never saw that comment hence why I was so confused.  Now I understand why you added the `requires triage`.

When we wrote up those rules it was not with a bot in mind but with a human process...that is, when we in the triage meeting saw `stale` + `verified` we would *choose* to remove verified if it needed re-verification.  But that is an action only humans can take (IMO), but that didn't really get reflected in the requirements.

The wording does not reflect that though, so that's definitely something we should change in the triage_process.md .

I would say the bot should just add the stale label, which we will see in triage and we handle what to do with the verified label at that time.

NickLaMuro:
> Can you explain this a bit more...

Sure.

Previously, this was one of the example stubs:

```ruby
double("stale_pr",
       :updated_at    => 7.months.ago,
       :fq_repo_name  => fq_repo_name,
       :number        => 2,
       :pull_request? => true,
       :labels        => [])
```

Specifically, the `labels` was stubbing the call to `.labels` directly, so that worked fine, but now that we have moved to using `GithubService::Issue#stale?` and `GithubService::Issue#.unmergeable?`, that no longer works since the `double` try and implement the logic for `#stale?` and `#unmergeable?` and use `:labels`.

So I had two options:

- Change the double to instead stub `#stale?` and `#unmergeable?` as booleans
- Do what I did here which instead effectively does the "`FactoryBot.build`" approach

I chose the latter route since it ends up exercising more of our codebase instead of hiding it behind stubs.

Unit test purists probably hate it since it is now testing outside the scope of this model, but I brought up `FactoryBot` (or `ActiveRecord::Base#new`) since it is pretty common practice to break that rule.  And since we only started testing any part of `GithubService` object (minus the `github_service/commands/) when I added the spec for it here:

https://github.com/ManageIQ/miq_bot/pull/481/files#diff-74e22cb2c056bfcfc22ba9006730335c

I figured it is probably better to actually test as far down the stack as possible instead of stubbing it away.

Also, since I am aware a rework of octokit is in the works:

https://github.com/octokit/octokit.rb/pull/1199#issuecomment-589970552

I think testing with as much of the internals of that gem is ideal so that when we inevitably upgrade that gem to the new release.  If we stub everything out, we don't catch anything with tests. 

> I saw in your other PRs you were going in this direction, and I don't understand why the need for something that appears more complicated at a glance.

This is different than probably what you are referring to, which is was the [`github_service_helper.rb`](https://github.com/ManageIQ/miq_bot/blob/4325a17/spec/support/github_service_helper.rb#L1) and that is not what I am using here.  Though, I probably could have, but decided to leave the existing tests as close to as they were as possible.

In this case, the specs already stubbed the fetch to github and returned the dummy objects from the stub.  If I were to have used `github_service_helper.rb`, the stub/dummy objects wouldn't have been necessary and only creating a stub response would have been required.

Functionally the amount of code that gets exercised is similar to what I did here, but two different approaches to doing that.  Again, was trying to reduce the diff of the original specs here to maintain their validity, so I chose not to do that as a result.

app/workers/stale_issue_marker.rb
NickLaMuro: Fixed

app/workers/stale_issue_marker.rb
NickLaMuro: Fixed

app/workers/stale_issue_marker.rb
NickLaMuro: Cuz GithubFlavoredMarkdown...

If you leave it as multi line:

```
Thank you for all your contributions!  More information about the ManageIQ
triage process can be found in [the traige process documentation][1].

[1]: https://www.manageiq.org/docs/guides/triage_process
```

**It will format like this:**

Thank you for all your contributions!  More information about the ManageIQ
triage process can be found in [the traige process documentation][1].



**Instead of like this:**

Thank you for all your contributions!  More information about the ManageIQ triage process can be found in [the traige process documentation][1].


(make note of the line break after "ManageIQ")


[1]: https://www.manageiq.org/docs/guides/triage_process

Fryguy:
Thanks for the info...yeah I was talking more about the github_service_helper, but also was curious about the change approach in general.

PR #485, opened 2020-03-12
Move unassignable and unremovable labels from constants to settings

bdunne:
These values really belong to the ManageIQ organization and the bot is intended to be reusable by people in other organizations.  Prefer settings for these types of values rather than hard-coded constants.

cc @NickLaMuro

### DIFF
--- a/lib/github_service/commands/add_label.rb
+++ b/lib/github_service/commands/add_label.rb
@@ -3,9 +3,9 @@ module GithubService
     class AddLabel < Base
       include IsTeamMember
 
-      UNASSIGNABLE = {
-        "jansa/yes" => "jansa/yes?"
-      }.freeze
+      def unassignable_labels
+        @unassignable_labels ||= Settings.labels.unassignable.to_h.stringify_keys
+      end
 
       private
 
@@ -57,7 +57,7 @@ module GithubService
 
       def handle_unassignable_labels(valid_labels)
         valid_labels.map! do |label|
-          UNASSIGNABLE.key?(label) ? UNASSIGNABLE[label] : label
+          unassignable_labels.key?(label) ? unassignable_labels[label] : label
         end
       end
 
--- a/lib/github_service/commands/remove_label.rb
+++ b/lib/github_service/commands/remove_label.rb
@@ -3,8 +3,6 @@ module GithubService
     class RemoveLabel < Base
       include IsTeamMember
 
-      UNREMOVABLE = %w[jansa/yes jansa/no].freeze
-
       alias_as 'rm_label'
 
       private
@@ -40,7 +38,7 @@ module GithubService
 
       def process_extracted_labels(issuer, valid_labels, _invalid_labels, unremovable)
         unless triage_member?(issuer)
-          valid_labels.each { |label| unremovable << label if UNREMOVABLE.include?(label) }
+          valid_labels.each { |label| unremovable << label if Settings.labels.unremovable.include?(label) }
           unremovable.each  { |label| valid_labels.delete(label) }
         end
       end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -18,6 +18,9 @@ bugzilla:
   product:
 grafana:
   url:
+labels:
+  unassignable: {}
+  unremovable: []
 
 # Worker settings
 diff_content_checker:
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -1,3 +1,9 @@
+:labels:
+  :unassignable:
+    jansa/yes: jansa/yes?
+  :unremovable:
+  - jansa/no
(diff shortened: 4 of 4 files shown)


lib/github_service/commands/remove_label.rb
```diff
@@ -3,7 +3,9 @@ module Commands
     class RemoveLabel < Base
       include IsTeamMember
 
-      UNREMOVABLE = %w[jansa/yes jansa/no].freeze
+      def unremovable_labels
```
NickLaMuro: Maybe make this a class variable so we aren't loading this on every instance?

lib/github_service/commands/add_label.rb
```diff
@@ -3,9 +3,9 @@ module Commands
     class AddLabel < Base
       include IsTeamMember
 
-      UNASSIGNABLE = {
-        "jansa/yes" => "jansa/yes?"
-      }.freeze
+      def unassignable_labels
```
NickLaMuro: Same here

lib/github_service/commands/remove_label.rb
NickLaMuro: Of course, if needed, you can alias it as an instance method, but I think that it should still be cached on the class level.

NickLaMuro:
@bdunne  I have a larger concern in general about over using `config/settings.yml` and making it less transparent about what is going on for those who don't run the bot in production, but I don't think that this PR is going against that concern and is only tangentially related.

Should I open an issue about this to get a discussion going?

lib/github_service/commands/add_label.rb
bdunne: It's already cached in `Settings` and that should get reloaded any time the settings are changed.  Are you concerned about the call to `stringify_keys`?

NickLaMuro: By that logic, then we just shouldn't be caching it at all, and makes me wonder why you have a `@` var there in the first place.

bdunne:
> Should I open an issue about this to get a discussion going?

Might be a good idea.  Currently the settings override on the box contains a mix of secrets and org-specific configurations.  I would be open to splitting the current config file into two files (one for secrets, one for everything else).  I'm not sure where we would put the new "everything else" file because I don't think that belongs in this repo.  @Fryguy Any thoughts?

lib/github_service/commands/add_label.rb
bdunne: I added the cache here so that we don't have to call `stringify_keys` for every label in the command.

bdunne: The other cache was not necessary, so I removed it.

NickLaMuro: Sorry, to be clear:

I don't really see this as a big performance hit, since we would be caching it for it's once use per instance of this (it seems), so I am more commenting on the fact that we are caching this on the instance, but it's cache should never be made use of since this is only ever called once per instance.

If we still want the cache, then probably `cache_with_timeout` would make sense based on the use case that you described:  changing the files and having it auto reload without a server restart.  That said, I have found that the `listen` gem has caused my CPU to spike to 100% on my current machine, so I tend to avoid trying to rely on that functionality.

NickLaMuro: > I added the cache here so that we don't have to call stringify_keys for every label in the command.

Correction:  You are right with this since it is being called with in a loop.  I would argue that it is negligible with the number of labels we currently have in this list and the number of labels that are passed in for `add_label` in general (one or two), and there is more savings to be had by caching this on a class level is that is your true intent.

Fryguy: > That said, I have found that the listen gem has caused my CPU to spike to 100% on my current machine, so I tend to avoid trying to relying on that gem/functionality.

Can you open an issue for this...I'd not heard this before and I don't have this issue nor do we see it in prod.

Fryguy:
> Should I open an issue about this to get a discussion going?

That sounds good to me.

The general idea is that while this is called the miq_bot, it's designed more generally than that (or at least should be).  I've had aspirations to move this code to a more generic bot system that others could use.

lib/github_service/commands/add_label.rb
Fryguy: I think the memoization is fine since this is at an instance level.  The next instance of AddLabel will query the Settings again.

NickLaMuro:
Opened said issue here:

https://github.com/ManageIQ/miq_bot/issues/486

Though, doesn't quite address the thoughts expressed here:

> The general idea is that while this is called the miq_bot, it's designed more generally than that (or at least should be). I've had aspirations to move this code to a more generic bot system that others could use.

I don't mind if the conversation there expands to design ideas on how we would approach something like that.

PR #494, opened 2020-04-01
[CommitMonitor] Send commit .details_hash instead of .full_message

NickLaMuro:
This is a followup to the changes from https://github.com/ManageIQ/miq_bot/pull/493 which ensure that we are only parsing the commit "message", and not the contents of the committer email, author email, etc.

A bit of a discussion on this can be found here:

https://gitter.im/ManageIQ/miq_bot?at=5e83d09b74c36c7bc678ff13

The PR itself is mostly a exercising in refactoring, and currently only the last two commits changes anything for the `CommitMonitor` and nested handler workers from that.

Notes on Usage of `new_commits_details`
---------------------------------------

It was brought up as a concern (by myself), that `new_commits_details` is potentially used by other handlers.  This is not the case, so this change is safe to make, but for background:

The two other `pull_request_monitor_handlers/` besides `CommitMetadataChecker`:

- `DiffContentChecker`:
  https://github.com/ManageIQ/miq_bot/blob/2a2ab16b3c094c3b65242cbee337172bbbf41834/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker.rb#L11
- `DiffFilenameChecker`:
  https://github.com/ManageIQ/miq_bot/blob/master/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_filename_checker.rb#L11

Both ignore `_new_commits_details`.

In addition, their are only two `CommitMonitor` handler groups that make use of `new_commits_details`:

- "commit" handlers (which there are none):
  https://github.com/ManageIQ/miq_bot/blob/9595dd1064ace4eef1ce835481b6a8b0af44e4d4/app/workers/commit_monitor.rb#L201-L208
- "batch" handlers (which only includes `GithubPrCommenter`)
  https://github.com/ManageIQ/miq_bot/blob/9595dd1064ace4eef1ce835481b6a8b0af44e4d4/app/workers/commit_monitor.rb#L226-L231

So it should be safe to make this change and no other class should miss it.

### DIFF
--- a/lib/git_service/commit.rb
+++ b/lib/git_service/commit.rb
@@ -70,5 +70,21 @@ module GitService
       message << "\n #{diff.status_summary}"
       message
     end
+
+    def details_hash
+      {
+        "sha"            => commit_oid,
+        "parent_oids"    => parent_oids,
+        "merge_commit?"  => parent_oids.length > 1,
+        "author"         => formatted_author,
+        "author_date"    => formatted_author_date,
+        "commit"         => formatted_committer,
+        "commit_date"    => formatted_committer_date,
+        "message"        => formatted_commit_message,
+        "files"          => diff.file_status.keys,
+        "stats"          => formatted_commit_stats,
+        "status_summary" => diff.status_summary
+      }
+    end
   end
 end
--- a/lib/git_service/commit.rb
+++ b/lib/git_service/commit.rb
@@ -23,13 +23,21 @@ module GitService
       other_commit.diff(rugged_commit)
     end
 
+    def formatted_author
+      "#{rugged_commit.author[:name]} <#{rugged_commit.author[:email]}>"
+    end
+
+    def formatted_author_date
+      rugged_commit.author[:time].to_time.strftime("%c %z")
+    end
+
     def full_message
       message = "commit #{commit_oid}\n"
       message << "Merge: #{parent_oids.join(" ")}\n" if parent_oids.length > 1
-      message << "Author:     #{rugged_commit.author[:name]} <#{rugged_commit.author[:email]}>\n"
-      message << "AuthorDate: #{rugged_commit.author[:time].to_time.strftime("%c %z")}\n"
-      message << "Commit:     #{rugged_commit.author[:name]} <#{rugged_commit.author[:email]}>\n"
-      message << "CommitDate: #{rugged_commit.author[:time].to_time.strftime("%c %z")}\n"
+      message << "Author:     #{formatted_author}\n"
+      message << "AuthorDate: #{formatted_author_date}\n"
+      message << "Commit:     #{formatted_author}\n"
+      message << "CommitDate: #{formatted_author_date}\n"
       message << "\n"
       message << rugged_commit.message.indent(4)
       message << "\n"
--- a/lib/git_service/commit.rb
+++ b/lib/git_service/commit.rb
@@ -39,6 +39,11 @@ module GitService
       rugged_commit.committer[:time].to_time.strftime("%c %z")
     end
 
+    # Note:  really not needed, but keeps it consistent
+    def formatted_commit_message
+      rugged_commit.message
+    end
+
     def formatted_commit_stats
       diff.file_status.map do |file, stats|
         file_stats  = file.dup
@@ -59,7 +64,7 @@ module GitService
       message << "Commit:     #{formatted_committer}\n"
       message << "CommitDate: #{formatted_committer_date}\n"
       message << "\n"
-      message << rugged_commit.message.indent(4)
+      message << formatted_commit_message.indent(4)
       message << "\n"
       message << formatted_commit_stats.join("\n").indent(1)
       message << "\n #{diff.status_summary}"
--- a/lib/git_service/commit.rb
+++ b/lib/git_service/commit.rb
@@ -31,13 +31,21 @@ module GitService
       rugged_commit.author[:time].to_time.strftime("%c %z")
     end
 
+    def formatted_committer
+      "#{rugged_commit.committer[:name]} <#{rugged_commit.committer[:email]}>"
+    end
+
+    def formatted_committer_date
+      rugged_commit.committer[:time].to_time.strftime("%c %z")
+    end
+
     def full_message
       message = "commit #{commit_oid}\n"
       message << "Merge: #{parent_oids.join(" ")}\n" if parent_oids.length > 1
       message << "Author:     #{formatted_author}\n"
       message << "AuthorDate: #{formatted_author_date}\n"
-      message << "Commit:     #{formatted_author}\n"
-      message << "CommitDate: #{formatted_author_date}\n"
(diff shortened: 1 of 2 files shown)


lib/git_service/commit.rb
```diff
@@ -23,21 +23,63 @@ def rugged_diff(other_ref = parent_oids.first)
...
 
+    def formatted_author
+      "#{rugged_commit.author[:name]} <#{rugged_commit.author[:email]}>"
+    end
+
+    def formatted_author_date
+      rugged_commit.author[:time].to_time.strftime("%c %z")
+    end
+
+    def formatted_committer
```
NickLaMuro, line 34: I mentioned in my commit message why I changed this here:

> [GitService::Commit] Fix Commit/CommitDate
> 
> In full_message, which is equivalent to the following `git-log` call:
>
>     $ git log -n1 --pretty=fuller --stat
>
> Has incorrectly defined `Commit` and `CommitDate` since it was added in
> commit 49b2146 of `ManageIQ/miq_bot`.
> 
> This addresses that, and handles `committer` in the same way that
> `author` has been formatted as well.

lib/git_service/commit.rb
```diff
@@ -23,21 +23,63 @@ def rugged_diff(other_ref = parent_oids.first)
...
+      "#{rugged_commit.committer[:name]} <#{rugged_commit.committer[:email]}>"
+    end
+
+    def formatted_committer_date
+      rugged_commit.committer[:time].to_time.strftime("%c %z")
+    end
+
+    def formatted_commit_stats
+      diff.file_status.map do |file, stats|
+        file_stats  = file
```
Fryguy, line 44: Note that the following line modify the file object directly...we may want this line to be `files_stats = file.dup`

lib/git_service/commit.rb
```diff
@@ -23,21 +23,63 @@ def rugged_diff(other_ref = parent_oids.first)
...
+    def details_hash
+      {
+        "sha"            => commit_oid,
+        "parent_oids"    => parent_oids,
+        "merge_commit?"  => parent_oids.length > 1,
+        "author"         => formatted_author,
+        "author_date"    => formatted_author_date,
+        "commit"         => formatted_committer,
+        "commit_date"    => formatted_committer_date,
+        "message"        => rugged_commit.message,
```
Fryguy, line 78: Should rugged_commit.message be moved out to a help method as well since it's also used within full_message?

lib/git_service/commit.rb
```diff
@@ -7,7 +7,7 @@ def initialize(rugged_repo, commit_oid)
     end
 
     def diff(other_ref = parent_oids.first)
-      Diff.new(rugged_diff(other_ref))
+      @diff ||= Diff.new(rugged_diff(other_ref))
```
Fryguy, line 10: This should not be memoized because the diff is dependent on the paramter

Fryguy: Or if it is memoized, then we need some sort of Hash by `other_ref`

lib/git_service/commit.rb
```diff
@@ -23,21 +23,63 @@ def rugged_diff(other_ref = parent_oids.first)
...
+    end
+
+    def formatted_commit_stats
+      diff.file_status.map do |file, stats|
+        file_stats  = file
+        file_stats << " | "
+        file_stats << (stats[:additions].to_i + stats[:deletions].to_i).to_s
+        file_stats << " "
+        file_stats << "+" if stats[:additions].positive?
+        file_stats << "-" if stats[:deletions].positive?
```
NickLaMuro, line 49: I did in fact go down the rabbit hole of looking into this:

https://github.com/ManageIQ/miq_bot/pull/398#discussion_r162480938

And for the time being, decided to eject from trying to handle that at the moment (I was digging into the internals of `git` for a bit before I realized I wasn't finding what I was looking for).  Effectively, the formatting that is done is based on the `width` available in `git log --stat[=width]` (where `width` is optional, and defaults to the width of the current terminal), so a decent amount of considerations are made to balance out the fields, and the "column sizes" (file path, total changes, additions/deletions graph) can vary based on the contents of each.

A couple of interesting commits from this repo (viewed by running `git log --stat=80 [SHA]` highlight some of the different scenarios based on the width:

- 7258ed4fc755a8ce3eca3249902bc38475f075cc
- b7916ca5c25f1b2bfd5d423b2086aa2cd4d979b0
- b327fd06e6db992b2ec332e92e9327d239a50a73
- a6f54c26aa9d099ec6cf01702b5c9f784093a83d

If we decide to attempt that at some point, the way I have broken out the method here should make it a little more readable to do just that.

lib/git_service/commit.rb
Fryguy: Oh yeah that's a definite bug in the old one.

lib/git_service/commit.rb
Fryguy: Yeah, we noticed that in the original PR and also bailed on it, cause it wasn't really that important to be *exactly* like git in that respect.

lib/git_service/commit.rb
NickLaMuro: I guess, though I guess I figured it wasn't really worth it compared to the others which had a bit of formatting logic attached.

lib/git_service/commit.rb
NickLaMuro: Hmmm, then I guess this was happening previously as well.  I will look into it.

lib/git_service/commit.rb
NickLaMuro: Yeah, that is fair.  Maybe I will just locally memoize then (or not bother and drop that commit), but basically `full_message` and `details_hash` call this method twice, so was trying to be a good "performance developer".

We will see, but I might just lean towards dropping it based no your comment.

lib/git_service/commit.rb
NickLaMuro: Yeah, mostly wanted to provide some context in this PR that "research was done", but ended up not going forward, since it wasn't completely clear why we bailed previously.

lib/git_service/commit.rb
NickLaMuro: Fixed now.

lib/git_service/commit.rb
NickLaMuro: Decided to do this.  There is a separate commit for the addition.

lib/git_service/commit.rb
NickLaMuro: Dropped.  Resolving.

NickLaMuro:
Note about the above `rubocop`s:

- Not fixing the date stuff since that is how it was, and didn't want to change the input/output of these methods, just the internal structure
- The last two are just without how I am aligning the `<<` and `=` when stacked.  If it really bothers someone, I can change it, but I would argue this form is easier to parse (and I also don't think rubocop is equipped to handle this specific use case... unsure)

Issue #496, opened 2020-05-05
Upgrade rubocop/haml-lint


NickLaMuro:
Based on some conversation here:

https://gitter.im/ManageIQ/miq_bot?at=5eb113dfb6dd230697a77f89

We are pretty out of date with `rubocop`, so this is probably worth while to do regardless of the `haml-lint` issues.

This probably isn't the only location that some changes need to be made (config changes probably need to happen in the ManageIQ/guides repo), but creating a tracking issue so we can keep track of the effort.

NickLaMuro:
not sure if the bot looks at this repo... but...

@miq-bot assign @NickLaMuro 
@miq-bot add_label dependencies

Fryguy:
Closing this as we rely on the rubocop updates from manageiq-style repo.  That repo can still get out of date, but this way it's centralized to one place.

PR #498, opened 2020-05-05
Upgrade rubocop to v0.82.0 and haml_lint to v0.35.0

NickLaMuro:
Requires `.rubocop_base.yml` changes found here:  https://github.com/ManageIQ/guides/pull/421

Further info and rational on the changes to the `.rubocop.yml` can be found in the commits for that repo.

Links
-----

- Partially addresses https://github.com/ManageIQ/miq_bot/issues/496
- Required merge in guides repo:  https://github.com/ManageIQ/guides/pull/421

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,8 +45,9 @@ gem 'awesome_spawn',        '>= 1.4.1'
 gem 'default_value_for',    '>= 3.1.0'
 gem 'haml_lint',            '~> 0.28.0', :require => false
 gem 'more_core_extensions', '~> 4.0.0',  :require => 'more_core_extensions/all'
-gem 'rubocop',              '~> 0.69.0', :require => false
-gem 'rubocop-performance',  '~> 1.3',    :require => false
+gem 'rubocop',              '~> 0.82.0', :require => false
+gem 'rubocop-performance',               :require => false
+gem 'rubocop-rails',                     :require => false
 gem 'rugged',                            :require => false
 
 gem 'octokit', '~> 4.8.0', :require => false
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,8 @@ gem 'travis',         '~> 1.7.6'
 
 gem 'awesome_spawn',        '>= 1.4.1'
 gem 'default_value_for',    '>= 3.1.0'
-gem 'haml_lint',            '~> 0.28.0', :require => false
+gem 'haml',                 '~> 5.1',    :require => false # force newer version of haml
+gem 'haml_lint',            '~> 0.35.0', :require => false
 gem 'more_core_extensions', '~> 4.0.0',  :require => 'more_core_extensions/all'
 gem 'rubocop',              '~> 0.82.0', :require => false
 gem 'rubocop-performance',               :require => false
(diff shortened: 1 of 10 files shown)


NickLaMuro:
Will rebase this before merge, but just wanted to get the specs passing first.

spec/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder_spec.rb
```diff
@@ -22,12 +22,12 @@
...
-- [ ] :exclamation: - [Line 4](https://github.com/some_user/some_repo/blob/8942a195a0bfa69ceb82c020c60565408cb46d3e/spec/workers/commit_monitor_handlers/commit_range/rubocop_checker/data/#{rubocop_check_directory}/coding_convention.rb#L4), Col 5 - [Layout/AlignHash](http://rubydoc.info/gems/rubocop/#{rubocop_version}/RuboCop/Cop/Layout/AlignHash) - Align the elements of a hash literal if they span more than one line.
+- [ ] :exclamation: - [Line 3](https://github.com/some_user/some_repo/blob/8942a195a0bfa69ceb82c020c60565408cb46d3e/spec/workers/commit_monitor_handlers/commit_range/rubocop_checker/data/#{rubocop_check_directory}/coding_convention.rb#L3), Col 5 - [Layout/HashAlignment](http://rubydoc.info/gems/rubocop/#{rubocop_version}/RuboCop/Cop/Layout/HashAlignment) - Align the keys and values of a hash literal if they span more than one line.
```
NickLaMuro, line 25: Note:  This needed to be regenerated because the class was changed from `AlignHash` to `HashAlignment`, and the `COP_URIS` constant:

https://github.com/ManageIQ/miq_bot/blob/995fc9fb25c46066dfe48d09a7e1eb31d1aebeef/app/workers/commit_monitor_handlers/commit_range/rubocop_checker/message_builder.rb#L32-L38

in the `MessageBuilder` as not correct.

Regenerated the example results for the specs as described here:

https://github.com/ManageIQ/miq_bot/blob/995fc9fb25c46066dfe48d09a7e1eb31d1aebeef/spec/support/rubocop_spec_helper.rb#L1-L14

locally with the config file changes for https://github.com/ManageIQ/guides/pull/421 applied.

Gemfile
```diff
@@ -43,10 +43,12 @@ gem 'travis',         '~> 1.7.6'
 
 gem 'awesome_spawn',        '>= 1.4.1'
 gem 'default_value_for',    '>= 3.1.0'
-gem 'haml_lint',            '~> 0.28.0', :require => false
+gem 'haml',                 '~> 5.1',    :require => false # force newer version of haml
```
NickLaMuro, line 46: This is just to make this Rails 5.2 compatible.  Probably not "really" required, but decided to put it in since `haml_lint` has a very loose `haml` dependency.

skateman: Well, if this gets pulled in as a dep into miq in development, this might be a problem. I'd feel safer with a `<5` here. cc @himdel

NickLaMuro: We can't do that, for the exact reasons why you requested this be upgraded in the first place:  `haml < 5` is not Rails 5.2 compatible.

- https://github.com/haml/haml/issues/895#issuecomment-282551500
- https://github.com/jrafanie/manageiq/commit/0ae87ac93f041ad955914b4e640b2c42c2dc725f

So if you were to use `miq_bot` in a `.gemspec`, `haml > 5` would be required.

Fryguy: I don't think we will include miq_bot in a gemspec, since it's not a gem, so I don't think there's an issue here.

Fryguy:
Merged.  @bdunne we need to coordinate deployment with merging ManageIQ/guides#421

PR #501, opened 2020-05-08
GitHub Status API Integration (rebased)

NickLaMuro:
This is a rebase of https://github.com/ManageIQ/miq_bot/pull/412 

This only required the update of `spec/lib/github_service_spec.rb` to be modified to work with the new `GithubService` specs that were added back in 1173807a to support `#username_lookup` (since both commits added this file back).

See #412 for details on this feature.


NickLaMuro:
@europ FYI:  This is just a rebase of your work.  If you want to pull it into your branch, that would work too and I can close this one.

Same goes for https://github.com/ManageIQ/miq_bot/pull/500 (though, that one is a bit more complex with it's changes and rational)

PR #502, opened 2020-05-11
Add Vagrantfile

NickLaMuro:
A basis for running miqbot via a Vagrant box instead of locally.

Probably needs some tweaks, but this is a first pass on getting it to bundle.


Fryguy:
@bdunne Please also review.  I merged anyway as we can iterate on this.  Just curious if you see anything that doesn't line up with prod.

PR #503, opened 2020-05-12
Upgrade ruby in .travis.yml and Vagrantfile

NickLaMuro:
`ruby-install` was originally picked for the `Vagrantfile` to match what was in the `.travis.yml`, and `dnf` only had 2.5.7 available.  But based on output from the bot in production:

> Checked commit NickLaMuro@b68ddde with ruby 2.5.7 ...
>
> Ref: https://github.com/ManageIQ/miq_bot/pull/497#issuecomment-624154539

We are using 2.5.7 there as well, so switching to use a `dnf` install to speed up the the over all build time of the box.

Also updating the `.travis.yml` to use 2.5.7 as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@
 
 # Ignore local seeds file
 db/seeds.local.rb
+
+# Ignore vagrant configuration/runtime files
+.vagrant/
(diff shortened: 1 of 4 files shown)


NickLaMuro:
Note:  Added some `Vagrantfile`/`.gitignore` updates to this PR that are not relevant to the original scope of the PR, but were forgotten in the original `Vagrantfile` PR.  Can move to another PR if desired, but figured it wasn't to big of a deal to add them here.

PR #505, opened 2020-05-19
GitService::Repo.git_fetch (and friends)

NickLaMuro:
This PR does a bunch of things to try and allow `.git_fetch` to work from `Rugged`:

- Moves logic for `Rugged` based `.git_fetch` into `GitService::Repo`
- Adds a `GitService::Credentials` in charge of finding credentials for a given remote/user
- Adds logic in `GitService::Credentials` to allow configuration of keys by host ~~(similar to `~/.ssh/config`)~~ using `~/.ssh/config` via the `net-ssh` gem
- Uses new `.git_fetch` and credential finding in `Repo` model
- Initializes git credentials (if configured) on Rails boot

Example config
---------------

~~This is what I am using for my ssh configuration for in my `config/settings/development.local.yml`:~~

**NOT NEEDED/USED.  Use example `~/.ssh/config` below instead:**

```yaml
git_service_hostname_ssh_config:
  '*':
    username: git
    privatekey: /full/path/to/.ssh/id_rsa
```

~~Full paths are required, it seems, and it will bomb when using `~/.ssh/id_rsa`.~~

This is what is in my `~/.ssh/config` for testing this:

```config
Host *
  AddKeysToAgent yes
  UseKeychain no
  IdentityFile ~/.ssh/id_rsa
```

(note:  I think `AddKeysToAgent` probably could be set to `no`, but I haven't tested)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem 'slim'
 
 # Services gems
 gem 'minigit',        '~> 0.0.4'
+gem 'net-ssh',        '~> 4.2.0'
 gem 'tracker_api',    '~> 1.6'
 gem 'travis',         '~> 1.7.6'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -206,6 +206,7 @@ GEM
       ruby2_keywords (~> 0.0.1)
     net-http-persistent (2.9.4)
     net-http-pipeline (1.0.1)
+    net-ssh (4.2.0)
     nio4r (2.5.2)
     nokogiri (1.10.8)
       mini_portile2 (~> 2.4.0)
@@ -415,6 +416,7 @@ DEPENDENCIES
   listen
   minigit (~> 0.0.4)
   more_core_extensions (~> 4.0.0)
+  net-ssh (~> 4.2.0)
   octokit (~> 4.8.0)
   pg
   rails (~> 5.2.2)
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -145,20 +145,8 @@ class Repo < ActiveRecord::Base
     results
   end
 
-  # TODO: Move this to GitService::Repo
   def git_fetch
-    # TODO: Figure out why this doesn't work on the bot using rugged
-    with_git_service { |git| git.fetch }
-    # require 'rugged'
-    # rugged_repo = Rugged::Repository.new(path.to_s)
-    # rugged_repo.remotes.each do |remote|
-    #   fetch_options = {}
-
-    #   username = extract_username_from_git_remote_url(remote.url)
-    #   fetch_options[:credentials] = Rugged::Credentials::SshKeyFromAgent.new(:username => username) if username
-
-    #   rugged_repo.fetch(remote.name, fetch_options)
-    # end
+    git_service.git_fetch
   end
 
   def create_branch!(branch_name)
--- a/lib/git_service/credentials.rb
+++ b/lib/git_service/credentials.rb
@@ -1,5 +1,13 @@
 module GitService
   class Credentials
+    # Generic method for finding git credentials based on what is available
+    #
+    # Will use ssh_agent if all other options have been exhausted.
+    #
+    def self.find_for_user_and_host(username, hostname)
+      from_ssh_agent(username)
+    end
+
     def self.from_ssh_agent(username)
       Rugged::Credentials::SshKeyFromAgent.new(:username => username)
     end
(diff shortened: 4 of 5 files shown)


NickLaMuro:
Of note:  I considered pulling in some code from `net/ssh` to parse a `~/.ssh/config` file as well:

https://github.com/NickLaMuro/miq_bot/tree/git-service-git-fetch-w-ssh-config

(last commit)

But decided against it since it was a lot of extra code, and most likely one of two options that have been provided here would be used anyway, so I scraped it.  Left the code in that branch those if you are curious.

Fryguy:
> Of note: I considered pulling in some code from net/ssh to parse a ~/.ssh/config file as well:
> But decided against it since it was a lot of extra code

It seems in the commit that you literally copied the code, but was it not accessible directly from net-ssh?

Fryguy:
Is the git host_config code necessary or can we just enfoce that ssh-agent is run for the bot?  What happens if ssh-agent is not running and `from_ssh_agent` is executed?

NickLaMuro:
> Is the git host_config code necessary or can we just enforce that ssh-agent is run for the bot?

It definitely isn't required to have the custom code for just using `Rugged::Credentials::SshKey`, though I knew specifically that that auth approach worked consistently since we use it in `ManageIQ/manageiq` for the `ansible/runner` code.

> What happens if ssh-agent is not running and from_ssh_agent is executed?

It seems like `Rugged::Credentials::SshKeyFromAgent` seems to be lazy loaded, so initializing this in `from_ssh_agent` should probably only happen `if username` exists in the uri, which I forgot to keep that code, so I will fix that.

Fryguy:
We use the host config code like that in manageiq?

NickLaMuro:
> It seems in the commit that you literally copied the code, but was it not accessible directly from net-ssh?

I figured adding that full gem for the use of literally one file was overkill, and since we didn't have the gem used in this project, it was a "lighter" solution.

> EDIT: Looking at the gem I can't see why we couldn't include that gem and then Net::SSH::Config.load("~/.ssh/config"). Then that would eliminate the entire custom host_config section, right?

Yup, that would definitely be possible.  That said, the end result would be similar to what you are doing with the `config/settings.yml` approach anyway, so I figured if you didn't already have that in place, it wouldn't be worth including a new gem to do something that a few lines of slightly custom code can do instead.

NickLaMuro:
> We use the host config code like that in manageiq?

No, I was saying that we use `SshKey` in `manageiq` (not `SshKeyFromAgent`).

NickLaMuro:
Also also:  Another reason I am doing this is because the Pronto Integration work currently does some custom `git_fetch` calls, and I would like this to be centralized in the `GitService` if possible.

Fryguy:
> so I figured if you didn't already have a ~/.ssh/config

We already do.  That's why I think I prefer the gem approach over extra config which would then be a duplicate of what's already in our ssh config.

NickLaMuro:
> That's why I think I prefer the gem approach over extra config which would then be a duplicate of what's already in our ssh config.

Okay, just keep in mind that while that parses it, I will still have to then convert the keys from that into the hash that `Rugged::Credentials::SshKey` understands.  The config above already does that, so I don't be surprised if the "gem approach" leads to more code still.

NickLaMuro:
@Fryguy so I updated my alternate branches to include two different approaches for the `net-ssh` code:

https://github.com/NickLaMuro/miq_bot/tree/git-service-git-fetch-w-ssh-config
https://github.com/NickLaMuro/miq_bot/tree/git-service-git-fetch-w-net-ssh

The first is what I had previously, but trims down the ported file to only the bits we need (with a few tweaks), and integrates it into `GitService::Credentials`.  The second instead uses the gem to do the same thing, and makes the necessary changes git `GitService::Credentials` to include it in the format we need it for `Rugged`.

Knowing you, I think I know which approach you would prefer, but decided to offer up for comparison.  Can integrate which ever you would rather see, and can remove some functionality from this branch if we decide we don't want the "`config/settings.yml` approach" now that we are parsing `~/.ssh/config`.

Fryguy:
I prefer the second approach in https://github.com/NickLaMuro/miq_bot/tree/git-service-git-fetch-w-net-ssh along with the removal of the config/settings bit.

NickLaMuro:
@Fryguy Okay, changes applied as requested.

Fryguy:
Lgtm, @bdunne can you also review?

Fryguy:
Oh there's one rubocop @NickLaMuro

NickLaMuro:
@Fryguy already fixed it, the bot just hasn't rechecked after my rebase.

NickLaMuro:
Now it has 😄

PR #508, opened 2020-06-11
[stale_issue_marker.rb] Fix 'updated' search

NickLaMuro:
`updated:<` should be the keyword used, not `update:<`, as shown by the difference in these two searches:

**Returns no records**

https://github.com/search?q=org%3AManageIQ+is%3Aopen+archived%3Afalse+update%3A%3C2020-03-01+-label%3Apinned&type=Issues

**Returns (a bunch of) records**

https://github.com/search?q=org%3AManageIQ+is%3Aopen+archived%3Afalse+updated%3A%3C2020-03-01+-label%3Apinned&type=Issues

### DIFF
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -28,7 +28,7 @@ class StaleIssueMarker
   private
 
   def handle_newly_stale_issues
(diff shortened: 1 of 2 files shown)


NickLaMuro:
Quick note about the search above:  It isn't "exactly" what is used by the bot, since I don't have a list of the `enabled_repos` available to actually use, but my guess is it is still pretty close.

PR #516, opened 2020-08-18
Ensure the clone is always up to date before returning a GitService::Repo

bdunne:
In a podified deployment with the queues running in different pods and the pods each having their own clones of the repos, we can't depend on other jobs having already fetched for the clone.

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -161,6 +161,7 @@ class Repo < ActiveRecord::Base
   end
 
   def git_service
+    git_fetch
     GitService::Repo.new(self)
   end
 


app/models/repo.rb
```diff
@@ -161,6 +161,7 @@ def create_branch!(branch_name)
   end
 
   def git_service
+    git_fetch
```
NickLaMuro, line 164: Drive by comment...

Would an alternative for this just use a shared volume between worker pods so that we aren't doing a bunch of duplicate clones per worker?  Just thinking that cloning `ManageIQ` for every worker container seems... expensive.

But I could also be full of it, and that also doesn't really invalidate this change, just questioning the justification for it and if it could be solved in a different way.

Fryguy:
Hmmm... The basic workflow of the entire thing was the commit monitor runs first which ends up fetching all of the repos once on a schedule.  Then, the various checkers run on the already pulled repos, specifically so they don't all have to fetch. It can get very expensive to do the fetches this often.  I don't really like this change, but I'm struggling to find a right way to do it in pods.  (Aside, not loving podified 😆 )

bdunne:
I also don't like having multiple copies of all of the repos around and having to keep them up to date.

`ReadWriteMany` volumes are hard to find and for our case (DigitalOcean) the only option is `ReadWriteOnce` so it can't be attached to multiple pods.  I could try setting up a NFS server in a pod and using that NFS share in the other pods.

bdunne:
Resolved to put both containers in the same pod.  If that becomes a problem we'll re-evaluate.

PR #519, opened 2020-10-15
[WIP] Add .docker-compose.yml

NickLaMuro:
Implements `docker-compose` workflow for development of `miq_bot`.

Usage
-----

1. `$ mkdir ./.ssh`
2. `$ cp ~/.ssh/id_rsa ./.ssh/miq-bot` (or equivalent for your setup)
3. `$ touch config/initializers/01_sidekiq_config.rb`
4. `$ docker-compose up`

Some of the above is to support the current workflow with production, so could be refined or improved in the future.

### Optional Steps

- If you are using `docker-machine`, then ensure you have a `docker-machine` environment running and configured for your terminal prior to calling `docker-compose`.
- If you want to pre-populate your bin

Known Issues
------------

### Uses `REF`, not local copy

Currently, the following patch is required to `Dockerfile` to allow using source instead of a particular version.

```diff
diff --git a/Dockerfile b/Dockerfile
index 34f5ff2..70e5ef6 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,14 +26,18 @@ RUN dnf -y --disableplugin=subscription-manager install \
     dnf -y --disableplugin=subscription-manager upgrade && \
     dnf clean all

-RUN wget https://github.com/ManageIQ/miq_bot/archive/$REF.zip && \
-    unzip $REF.zip -d /opt && \
-    rm -rf $REF.zip && \
-    mv /opt/miq_bot-* $APP_ROOT && \
-    chgrp -R 0 $APP_ROOT && \
-    chmod -R g=u $APP_ROOT
+# RUN wget https://github.com/ManageIQ/miq_bot/archive/$REF.zip && \
+#     unzip $REF.zip -d /opt && \
+#     rm -rf $REF.zip && \
+#     mv /opt/miq_bot-* $APP_ROOT && \
+#     chgrp -R 0 $APP_ROOT && \
+#     chmod -R g=u $APP_ROOT

-RUN dnf -y --disableplugin=subscription-manager --setopt=tsflags=nodocs install \
+COPY . $APP_ROOT
+
+RUN chgrp -R 0 $APP_ROOT && \
+    chmod -R g=u $APP_ROOT && \
+    dnf -y --disableplugin=subscription-manager --setopt=tsflags=nodocs install \
       @development \
       cmake \
       git \
```

Not sure the best way to handle this in the future to support both production deploy and local development.

### Rebuild is slow

We probably could re-order the `Dockerfile` a bit to allow an easier rebuild of the `app` containers.  Didn't want to tackle that in a first pass.

### `rugged` isn't being built with `openssl`

Need to look into what is causing this.  Not sure if it is a issue with the setup or in production as well.


bdunne:
What is the advantage of doing this?  Don't you need to rebuild the container(s) with a single line change?  I'd prefer to stick with `foreman start` for the developer workflow.

NickLaMuro:
> Don't you need to rebuild the container(s) with a single line change?

You do, though, that is why I mentioned that in the PR description (see: "Rebuild is slow").  This could be solved most likely with multi-stage builds, but I haven't tested, nor do I think it make drastic changes to the `Dockerfile` in this PR for a first pass (hence me just sharing the `diff` above, not committing it).

But to answer the rest:

> What is the advantage of doing this?
> ...
>  I'd prefer to stick with foreman start for the developer workflow.

Fine.  You, personally, can still do that.  This isn't meant to replace anything, just providing an alternative option for running the application in development with as few steps to setup as possible, and attempt to mimic production as much as possible in doing so as well.

However, for the `eslint`/`scss-lint` (will probably be `style-lint` because [#reasons™](https://github.com/sds/scss-lint#notice-consider-other-tools-before-adopting-scss-lint)) effort:

https://github.com/ManageIQ/miq_bot/pull/221#issuecomment-708485734

I am going to need to:

* install `node` (currently not done, just the package exists)
* probably install yarn
* install said packages from above
* possibly manipulate the path to make sure those are available to the bot

And testing all of the locally with foreman is next to worthless if it never is going to be run a VM in production anyway.  An frankly, the idea of using 2+ gigs of memory just to boot `minishift` w/o containers does not sound appealing, so I put together something that was a little more light weight and stream lined for my development purposes.


* * *

That all said, I don't need this merged for me to use it.  If others don't find this useful, that is fine, but I probably will use this in my testing of these new changes.  Because having to rebuild the container every time is far better then having to do both that **AND** having to push a ref to GitHub **AND** run minikube for me personally.

NickLaMuro:
This probably needs to be rebased now that `/opt/opt` has been fixed in master:

https://github.com/ManageIQ/miq_bot/pull/520

PR #522, opened 2020-10-22
upgrade faraday

d-m-u:
This is the last reference to faraday < 1.0 in our org and it's about time to update.


Fryguy:
@bdunne Are we expecting the bot to fail on manageiq-style issues already?

bdunne:
> @bdunne Are we expecting the bot to fail on manageiq-style issues already?

Only on this repo.

d-m-u:
I'd also like to close this, since it isn't necessary.

PR #524, opened 2020-10-26
For cross-repo-test, make PR Name more specific

chessbyte:
This avoids each auto-generated PR to have the identical name of `[BOT] Cross repo test run`

Follow-up to https://github.com/ManageIQ/miq_bot/pull/454


NickLaMuro (approved):
@chessbyte Works for me.  I _**think**_ the only reason I didn't do that was to avoid duplicate info, since it is also in the PR, but I don't see that being a big issue.

👍

PR #526, opened 2020-11-02
Process repository groups in cross-repo-tests command

agrare:
Allow for users to specify groups of repositories with a simple `/group-name` directive.
These groups can be defined dynamically in manageiq-release config as yaml files.

Example: `@miq-bot cross_repo_tests /providers including manageiq#1234`

TODO
- [x] Define the repo groups yaml file in manageiq-release https://github.com/ManageIQ/manageiq-release/pull/169

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -279,7 +279,7 @@ module GithubService
         response = Net::HTTP.get_response(uri)
         return {} if response.code != 200
 
-        YAML.load(response.body).transform_values(&:keys)
+        YAML.safe_load(response.body, :aliases => true).transform_values(&:keys)
       end
       private_class_method :fetch_manageiq_release_repo_groups
 
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -75,6 +75,9 @@ module GithubService
 
       restrict_to :organization
 
+      # Cache the repo groups yaml file from https://github.com/ManageIQ/manageiq-release
+      cache_with_timeout(:repo_groups_hash, 60.minutes) { fetch_manageiq_release_repo_groups }
+
       def self.test_repo_url
         Settings.cross_repo_tests.url
       end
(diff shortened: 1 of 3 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -246,6 +268,18 @@ def create_cross_repo_test_pull_request
                                           "[BOT] Cross repo test for #{issue.fq_repo_name}##{issue.number}", pr_desc)
       end
 
+      def self.fetch_manageiq_release_repo_groups
+        require 'net/http'
+        require 'uri'
+
+        uri = URI.parse("https://raw.githubusercontent.com/ManageIQ/manageiq-release/master/config/repo_groups.yml")
```
Fryguy, line 275: To make it also usually in manageiq-release, let's name it repos.groups.yml

agrare: :+1:

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -75,6 +75,9 @@ class CrossRepoTest < Base
 
       restrict_to :organization
 
+      # Cache the repo groups yaml file from https://github.com/ManageIQ/manageiq-release
+      cache_with_timeout(:repo_groups_hash, 60.minutes) { fetch_manageiq_release_repo_groups }
```
Fryguy, line 79: I think it might be cleaner if the fetch method one-time expands the known list and just stores that as a hash.  Then later, it's a straight hash lookup, with not reiteration.

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -134,18 +137,37 @@ def parse_value(value)
...
       end
 
+      def expand_repo_groups(repo_groups)
+        # Expand out each group and prioritize any repos which were passed explicitly
+        # in the test_repos list by the user
+        #
+        # E.g. `cross-repo manageiq-providers-amazon#1234 /providers` should only
+        # include manageiq-providers-amazon#1234
+        repo_groups.flat_map { |repo_group| expand_repo_group(repo_group) }
+                   .reject { |repo_name| @test_repos.any? { |test_repo| test_repo.include?(repo_name) } }
```
Fryguy, line 162: I think you can remove the reject and leave this as a pure function.  Then up above do

```ruby
@test_repos = @test_repos | expand_repo_groups(repo_groups)
```

which will do a set union and eliminate the dups for you.

lib/github_service/commands/cross_repo_test.rb
agrare: You mean don't cache_with_timeout, just fetch once and cache it?

Fryguy: I think I misread the code, because I saw the expansion in a loop and though it was doing some more complicated expansion the list on each pass.

lib/github_service/commands/cross_repo_test.rb
agrare: So I used `.include?` instead of a straight `|=` or `.uniq` because most likely a test_repo passed in by the user is going to be a PR/branch/sha e.g. `@miq-bot cross_repo_tests manageiq-providers-amazon#1234 /providers` (see the spec test covering this case) and we want those to take precedence over the expanded repo groups

Fryguy: Oh I see...hmm then maybe we need a bit of an expanded comment as I didn't realize that was the problem you were trying to solve.  In any case, :+1:

agrare: I can add that we're looking for a substring in the test_repos list specifically because we want to prioritize a PR over the bare repo otherwise I think the comment covers what we're doing and not necessarily how it is being done

agrare: Actually I think I need to do a regex not just `.include?` because `manageiq-providers-azure` would also match on `manageiq-providers-azure_stack`

agrare: Okay this should pick up on `(ManageIQ/)manageiq-providers-azure(#|@)1234` but not `manageiq-providers-azure_stack` now, spec added for this case also

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -134,18 +137,40 @@ def parse_value(value)
         @value = value
 
         @test_repos, @repos = value.split(/\s+including\s+/)
-                                   .map { |repo_list| repo_list.split(",") }
+                                   .map { |repo_list| repo_list.split(",").map(&:strip) }
```
agrare, line 140: This was causing issues with multiple test_repos like `repo1, /providers` because the latter would be `" /providers"` and the `start_with?("/")` was failing, seemed better to just strip all leading/trailing whitespace anyway

djberg96:
This looks nice.

Can the directive take arguments? Something like e.g. `@miq-bot cross_repo_tests /providers amazon,azure,vmware including manageiq#1234`

agrare:
@djberg96 interesting I hadn't considered that, right now this only expands out the groups passed in but I was starting to think we could start doing regex expansion now that we have a cached list of "all" of the repos so something like `@miq-bot cross_repo_tests manageiq-providers*` could work

PR #530, opened 2020-11-09
Fix testing Net::HTTP response code

agrare:
We were testing that "200" != 200 and returning that we failed to fetch the repo sets yaml file, and .flat_map { {}["all"] } was resulting in the expanded repo groups being [nil]

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -162,6 +162,7 @@ module GithubService
         # include manageiq-providers-amazon#1234 not both manageiq-providers-amazon#1234
         # and manageiq-providers-amazon
         repo_groups.flat_map { |repo_group| expand_repo_group(repo_group) }
+                   .compact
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -277,7 +278,7 @@ def self.fetch_manageiq_release_repo_groups
 
         uri = URI.parse("https://raw.githubusercontent.com/ManageIQ/manageiq-release/master/config/repos.sets.yml")
         response = Net::HTTP.get_response(uri)
-        return {} if response.code != 200
+        return {} if response.value
```
Fryguy, line 281: I think this is wrong.  From what I read `value` will raise an exception if the error code is not 200.  But this code will return {} if it doesn't fail?

agrare: :facepalm: you're right updated to call `response.value` and rescue the exception and return `{}`

agrare:
Okay added test coverage for this method so I don't :facepalm: this again haha

PR #532, opened 2020-11-11
update i18n config fallbacks line

d-m-u:
I think this needs to be updated: 

Please check your Rails app for 'config.i18n.fallbacks = true'.
If you're using I18n (>= 1.1.0) and Rails (< 5.2.2), this should be
'config.i18n.fallbacks = [I18n.default_locale]'.
If not, fallbacks will be broken in your app by I18n 1.1.x.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -67,7 +67,7 @@ MiqBot::Application.configure do
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation can not be found).
-  config.i18n.fallbacks = true
+  config.i18n.fallbacks = [I18n.default_locale]
 
   # Send deprecation notices to registered listeners.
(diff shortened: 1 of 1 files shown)


chessbyte:
@d-m-u in the core [Gemfile](https://github.com/ManageIQ/manageiq/blob/master/Gemfile), `gem "rails", "~>5.2.4", ">=5.2.4.4"` is specified.  The message `If you're using I18n (>= 1.1.0) and Rails (< 5.2.2)` seems to imply that this is not applicable to ManageIQ because we use a Rails version greater than 5.2.2.  What am I missing?

Fryguy:
Core Gemfile is irrelevant here, since the bot is standalone, but we are on 5.2.2+ as per https://github.com/ManageIQ/miq_bot/blob/master/Gemfile#L4, so we don't need this change.

PR #534, opened 2021-01-18
Ensure test_repos that are branches/PRs are part of the repos list

Fryguy:
Example (assuming the PR in questions is `manageiq#1234`)

`@miq-bot cross_repo_test manageiq-api#2345, manageiq-ui-classic#3456` would produce:

```diff
-REPOS=ManageIQ/manageiq#1234
+REPOS=ManageIQ/manageiq#1234,ManageIQ/manageiq-api#2345,ManageIQ/manageiq-ui-classic#3456
 ...
 TEST_REPO=ManageIQ/manageiq-api#2345
 TEST_REPO=ManageIQ/manageiq-ui-classic#3456
```

---

The "Before" is not accurate because the tests are not running within the entire suite of changes as can be seen in the "After".  Thus, this change no longer puts the onus on the developer to remember to add everything (again) in the including section.

@agrare Please review.

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -49,11 +49,6 @@ module GithubService
     #   -  - TEST_REPO=
     #   +  - TEST_REPO=ManageIQ/manageiq-api
     #   +  - TEST_REPO=ManageIQ/manageiq-ui-classic#5678
-    #
-    # TODO:  Handle the "self" case, where `manageiq` is also a TEST_REPO
-    #
-    # (maybe include a "self" helper as well?)
-    #
     class CrossRepoTest < Base
       # Reference to the branch we are creating off of origin/master
       attr_reader :branch_ref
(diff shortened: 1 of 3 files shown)


agrare (approved):
Yeah this is more intuitive to me and I can't think of a case where you'd have a PR/branch that you want to test that you actively wouldn't want to be overridden for the other tests.

agrare:
@Fryguy maybe add a comment to the cross_repo_test command documentation in the README to indicate that anything in the list of test_repos will be overridden for the all the other tests as well?  It is a little ambiguous at the moment and the only "documentation" on the behavior is "Runs cross repo tests that include the current PR code base."

Fryguy:
@agrare Updated.  I also noticed the comment on the class itself _already_ expected to work this way (that is, this is effectively a bug fix since it was presumed to work this way), so I didn't need to update the comment

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -173,9 +175,14 @@ def expand_repo_group(repo_group)
       end
 
       def normalize_repo_name(repo)
+        repo = repo.strip
         repo.include?("/") ? repo : "#{issue.organization_name}/#{repo}"
       end
 
+      def branch_or_pr?(repo)
```
NickLaMuro, line 180: Seems like this method should have been added in the next commit... instead of part of this one.

Fryguy: Oh yeah you're right...I missed that on the rebase -i

PR #537, opened 2021-01-20
Add the PR identifier to the list of test repos

Fryguy:
@agrare Please review.

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -47,6 +47,7 @@ module GithubService
     #   +  - REPOS=Fryguy/more_core_extensions@feature,Fryguy/linux_admin@feature,ManageIQ/manageiq#1234,manageiq-ui-classic#5678
     #      matrix:
     #   -  - TEST_REPO=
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -139,8 +140,8 @@ def parse_value(value)
         repo_groups, @test_repos = @test_repos.partition { |repo| repo.start_with?("/") }
         @test_repos |= expand_repo_groups(repo_groups)
 
-        # Add the identifier for the PR for this comment to repos
-        @repos << "#{issue.repo_name}##{issue.number}"
+        # Add the PR for this comment to the list of test repos
+        @test_repos << "#{issue.repo_name}##{issue.number}"
 
         # Add any test_repos that are branches or PRs to the repos list
         @repos |= @test_repos.select { |repo| branch_or_pr?(repo) }
```
Fryguy, line 147: Note that since `@repos` is a super-set of `@test_repos` from here, we don't have to change the above to append to both and we only have to append to `@test_repos`

agrare (approved):

Issue #538, opened 2021-01-22
Current linter issues


djberg96:
Via `rubocop -l`:

```
Inspecting 192 files
........WW..........W......................WW.....................................W.....W.........WWW..........................W.................W.W.W.WW.................W...EWWWWW............

Offenses:

app/models/branch.rb:34:51: W: [Correctable] Lint/UnifiedInteger: Use Integer instead of Fixnum.
    n = MinigitService.pr_branch(n) if n.kind_of?(Fixnum)
                                                  ^^^^^^
app/models/repo.rb:189:3: W: Lint/SuppressedException: Do not suppress exceptions.
  rescue Errno::ENOENT
  ^^^^^^^^^^^^^^^^^^^^
app/workers/commit_monitor_handlers/commit_range/path_based_labeler.rb:27:3: W: Lint/SuppressedException: Do not suppress exceptions.
  rescue GitService::UnmergeableError
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
config/application.rb:27:5: W: Lint/Debugger: Remove debugger entry point console.
    console do
    ^^^^^^^
config/boot.rb:4:33: W: [Correctable] Lint/DeprecatedClassMethods: File.exists? is deprecated in favor of File.exist?.
require 'bundler/setup' if File.exists?(ENV['BUNDLE_GEMFILE'])
                                ^^^^^^^
lib/bugzilla_service.rb:148:3: W: Lint/IneffectiveAccessModifier: private (on line 121) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
  def self.ids_in_git_commit_message(message)
  ^^^
lib/bugzilla_service.rb:152:3: W: Lint/IneffectiveAccessModifier: private (on line 121) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
  def self.search_in_message(message)
  ^^^
lib/git_service/diff.rb:30:21: W: Lint/AssignmentInCondition: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
        if new_file = patch.delta.new_file.try(:[], :path)
                    ^
lib/git_service/diff.rb:34:21: W: Lint/AssignmentInCondition: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
        if old_file = patch.delta.old_file.try(:[], :path)
                    ^
[...]

PR #539, opened 2021-01-22
Ignore spec data in rubocop

Fryguy:
This data has intentional errors in it so that when the bot runs rubocop
it can pick up real errors (and thus report on them).

Partial fix for #538 (drops the number of offenses in this repo from 121
to 18)


Fryguy:
@bdunne Please review. cc @djberg96

djberg96:
👍🏻 LGTM

PR #541, opened 2021-01-26
Rubocop fixes

djberg96:
Just a small PR that addresses some linter issues brought up in https://github.com/ManageIQ/miq_bot/issues/538:

* Use Integer instead of Fixnum
* Remove unused variables
* Re-enable a cop when finished
* File.exist? vs File.exists?
* ~~Make private class methods private~~
* Remove unnecessary require

### DIFF
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -31,7 +31,7 @@ class Branch < ActiveRecord::Base
   end
 
   def self.with_branch_or_pr_number(n)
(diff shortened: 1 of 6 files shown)


Fryguy (approved):

Fryguy:
Seems we were testing those private methods, so you'll have to do .send

djberg96:
@Fryguy I just removed those for now.

NickLaMuro (approved):
These seem pretty minor to me.  👍

PR #544, opened 2021-02-05
Handle duplicate repo name where one is a PR and the other is bare

Fryguy:
This handles the case where someone asks for the same repo multiple times,
but where one is a PR and the other is the bare repo.  This can even happen
accidentally, such as in the following case:

In ManageIQ/manageiq PR 1234, someone asks the bot to process /all, but
/all includes ManageIQ/manageiq.  The bot adds the PR to the test repo
list and now you have both the PR and the bare repo.  Additionally, it
will add the PR to the included list. manageiq-cross-repo will then see
that as an invalid combination because you can't test a bare repo while
also saying that it should include a PR version of that same repo.

This commit will prioritize PRs over bare repos when duplicates are
detected.

@agrare Please review.

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -136,35 +136,33 @@ module GithubService
                                    .map { |repo_list| repo_list.split(",").map(&:strip) }
         @repos ||= []
 
-        # Expand repo groups (e.g. /providers)
-        repo_groups, @test_repos = @test_repos.partition { |repo| repo.start_with?("/") }
-        @test_repos |= expand_repo_groups(repo_groups)
+        # Ignore bare repos in include list
+        @repos.select! { |repo| branch_or_pr?(repo) }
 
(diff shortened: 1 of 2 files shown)


Fryguy:
Real world example where this happened - https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/295/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R21-R22

Fryguy:
I did not handle possible invalid cases in this PR.  I plan to do those in a follow up.

- If someone specifies two different PR numbers for the same repo (right now we just pick the last one after sorting)
- If someone specifies invalid repos entirely (these will just cause failures in the cross-repo-test PR itself
- If someone specifies bare repos in the including directive (these are just straight-up ignored right now)

agrare (approved):

agrare:
Yeah I like the idea of doing a sanity check on the args and having the bot comment back if it is invalid

PR #545, opened 2021-02-05
Validate invalid repo names and repo name conflicts

Fryguy:
Follow up to #544 

@agrare Please review.

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -187,7 +187,10 @@ module GithubService
           .sort # Unadorned repo name will always sort before adorned repo names
           .uniq
           .slice_when { |a, b| bare_repo_name(a) != bare_repo_name(b) }
-          .map(&:last)
+          .map do |repos|
+            repos.shift if repos.size > 1 && bare_repo?(repos.first)
+            repos.size == 1 ? repos.first : repos
+          end
       end
 
       def bare_repo_name(repo)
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -187,18 +187,31 @@ def dedup_repo_list(repo_list)
           .sort # Unadorned repo name will always sort before adorned repo names
           .uniq
           .slice_when { |a, b| bare_repo_name(a) != bare_repo_name(b) }
-          .map(&:last)
+          .map do |repos|
+            repos.shift if repos.size > 1 && bare_repo_name?(repos.first)
+            repos.size == 1 ? repos.first : repos
+          end
```
Fryguy, line 193: Previously this just took the last one even if there were conflicts.  Now it will allow the conflicts to bubble out, so that `.valid?` can inspect them.  Note that a `[repo, repo#PR]` is not conflict...we prioritize the PR, which is why we drop the bare repo.

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -207,6 +220,37 @@ def valid?
         true
       end
 
+      def validate_repo_names
+        invalid_repo_names = (@test_repos + @repos).select { |r| r.include?(" ") }.sort.uniq
```
Fryguy, line 224: For now I just check for a space, because that's the easiest way to handle certain invalid things like forgetting to use commas to separate repo, or using the word `includes` instead of `including`. Eventually we can fully check all valid characters if we wanted.

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -187,18 +187,31 @@ def dedup_repo_list(repo_list)
...
+            repos.shift if repos.size > 1 && bare_repo_name?(repos.first)
+            repos.size == 1 ? repos.first : repos
+          end
       end
 
       def bare_repo_name(repo)
         repo.split(/[@#]/, 2).first
       end
 
+      def bare_repo_name?(repo)
```
agrare, line 200: I wonder if `bare_repo?` would be a better name since this is the inverse of `branch_or_pr?` ?  It looks confusing that there is a `bare_repo_name` and a `bare_repo_name?` that are unrelated.

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -207,6 +220,37 @@ def valid?
...
+          message = "@#{issuer} 'cross-repo-test(s)' was given invalid repo names and cannot continue\n\n"
+          invalid_repo_names.each do |repo|
+            message << "* `#{repo}`\n"
+          end
+
+          issue.add_comment(message)
+          return false
+        end
+
+        true
```
agrare, line 235: Minor, but the flow could be a little cleaner if we return true early like
```
invalid_repo_names = (@test_repos + @repos).select { |r| r.include?(" ") }.sort.uniq
return true if invalid_repo_names.empty?

message = "@#{issuer} 'cr...
...

false
```

Fryguy: Oh yeah I like that

Fryguy:
@agrare Updated.

agrare (approved):

PR #546, opened 2021-02-17
Handle bare PR numbers in cross-repo-test

Fryguy:
For example,

If in PR `Foo/bar#1234`, if someone does, `@miq-bot cross-repo-test #2345`, then it will fail showing a conflict between `Foo/bar#1234` and `Foo/bar#2345`.  Note that a bare PR number will always conflict, so we could change this to fail-fast, but I thought the conflict error message would be much cleaner.

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -177,7 +177,9 @@ module GithubService
 
       def normalize_repo_name(repo)
         repo = repo.strip
-        repo.include?("/") ? repo : "#{issue.organization_name}/#{repo}"
+        repo = "#{issue.repo_name}#{repo}" if repo.start_with?("#")
+        repo = "#{issue.organization_name}/#{repo}" unless repo.include?("/")
+        repo
       end
 
       # Deduplicates entries sharing the same bare repo name, prioritizing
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -177,7 +177,9 @@ def normalize_repo_list(repo_list)
 
       def normalize_repo_name(repo)
         repo = repo.strip
-        repo.include?("/") ? repo : "#{issue.organization_name}/#{repo}"
+        repo = "#{issue.repo_name}#{repo}" if repo.start_with?("#")
```
agrare, line 180: If we're going to handle bare PR numbers should we also do the same for `start_with?("@")` for branches/shas?

Fryguy: No, because you can't specify a bare branch - that would be an `@-mention`

agrare: https://github.com/ManageIQ/manageiq/pull/21060#issuecomment-780668127
https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/323/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R21

Fryguy: Ok, I can add it in...it's really weird, but at least the normalization would be "normal"

agrare: Yeah its really just another case of "Repository doesn't exist" so maybe not even worth it

agrare (approved):

PR #548, opened 2021-02-22
Fix recursive bot commands

NickLaMuro:
Removes bot thinking it is calling itself when making a comment where it's username could be used when adding a label.

See the following for an example:

  https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/124#issuecomment-783647447

### DIFF
--- a/lib/github_service/command_dispatcher.rb
+++ b/lib/github_service/command_dispatcher.rb
@@ -26,6 +26,7 @@ module GithubService
       lines.each do |line|
         match = command_regex.match(line.strip)
         next unless match
+        next if issuer == bot_name
 
         command       = match[:command]
         command_value = match[:command_value]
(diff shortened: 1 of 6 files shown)


Fryguy:
Oh i see - this is label specific...I'm concerned this could happen with _any_ command where the bot replies

lib/github_service/command_dispatcher.rb
```diff
@@ -26,6 +26,7 @@ def dispatch!(issuer:, text:)
       lines.each do |line|
         match = command_regex.match(line.strip)
         next unless match
+        next if issuer == GithubService.bot_name
```
Fryguy, line 29: Interestingly we also have the bot_name method down below on line 55

NickLaMuro: Should we coalesce around a single location for getting that?

NickLaMuro: @Fryguy Alright, I made this consistent now.

NickLaMuro:
First commit trues up our usage of `bot_command`, and makes it a method on `GithubService`.  The rest of the PR is fixing the two spots previously identified... now with infinitely more tests than before (on the PR)!

PR #550, opened 2021-03-04
Add production rake tasks

Fryguy:
- `production:edit_settings` to modify the bot settings in production
- `production:console` to open a console in production
- `release` to tag, build, and release a new version into production

@bdunne Please review.

Note that I released a 0.13.7 with the bot using this

### DIFF
--- a/lib/tasks/production.rake
+++ b/lib/tasks/production.rake
@@ -1,0 +1,150 @@
+module Kubernetes
+  def self.available?
+    !`which kubectl`.chomp.empty?
+  end
+
+  def self.config_dir
+    @config_dir ||= Pathname.new("~/.kube").expand_path
+  end
+
+  def self.config_file
+    @config_file ||= config_dir.join("config")
+  end
+
+  def self.merge_config(file)
+    system({"KUBECONFIG" => "#{config_file}:#{file}"}, "kubectl config view --flatten > #{config_file}")
+  end
+
+  def self.context?(context)
+    !`kubectl config get-contexts --no-headers #{context} 2>/dev/null`.chomp.empty?
+  end
+
+  def self.current_namespace
+    `kubectl config view --minify --output jsonpath={.contexts..namespace}`.chomp
+  end
+
+  def self.use_context(context)
+    system("kubectl config use-context #{context}")
+  end
+
+  def self.use_namespace(namespace)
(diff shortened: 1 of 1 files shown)


lib/tasks/production.rake
```diff
@@ -0,0 +1,152 @@
...
+  else
+    puts "Already tagged '#{version}'"
+  end
+  exit 1 unless system("git push #{remote} #{version}")
+
+  image = "manageiq/miq_bot:#{version}"
+
+  puts
+  puts "Building docker image..."
+  exit 1 unless system("docker build . -t #{image}")
```
Fryguy, line 135: Not sure if it's preferable to do a `--no-cache` here or not.

bdunne: Yes, I usually do `--no-cache`

lib/tasks/production.rake
```diff
@@ -0,0 +1,152 @@
...
+    exit 1 unless system("git tag #{version}")
+  elsif `git tag --points-at HEAD`.chomp != version
+    $stderr.puts "ERROR: Already tagged '#{version}', but you are not on that commit"
+    exit 1
+  else
+    puts "Already tagged '#{version}'"
+  end
+  exit 1 unless system("git push #{remote} #{version}")
+
+  image = "manageiq/miq_bot:#{version}"
```
bdunne, line 131:
```suggestion
  image = "docker.io/manageiq/miq_bot:#{version}"
```

lib/tasks/production.rake
Fryguy: yeah it actually requires it, otherwise it doesn't detect a code change and uses the cache for the entire thing.

lib/tasks/production.rake
Fryguy: If i do a docker build with that will it still work?  I had it separate because I wasn't sure if the docker build step could handle that

lib/tasks/production.rake
Fryguy: Added

lib/tasks/production.rake
Fryguy: It worked...Updated

Fryguy:
@bdunne Updated with your changes

PR #560, opened 2021-04-27
Update manageiq-style, including rubocop update

Fryguy:


spec/workers/commit_monitor_handlers/commit_range/rubocop_checker/data/with_results_with_no_offenses/no_offenses.rb
```diff
@@ -1,2 +1,5 @@
 class NoOffenses
+  def initialize
+    @x = 1
+  end
```
Fryguy, line 4: This was needed because of https://docs.rubocop.org/rubocop/1.13/cops_lint.html#lintemptyclass

spec/workers/commit_monitor_handlers/commit_range/rubocop_checker/data/with_results_with_offenses/ruby_syntax_error.rb
```diff
@@ -1,3 +1,6 @@
 class RubySyntaxError
+  def initialize
+    @x = 1
+  end
```
Fryguy, line 4: This technically wasn't needed, but since I did it above to avoid the EmptyClass, I did it here too.

Fryguy:
Note that the cop URLs are kind of messed up because not all cops are under rubocop gem anymore.  I'm looking into replacing that with pointing to https://docs.rubocop.org

PR #561, opened 2021-04-27
Use RuboCop::Cop::Base#documentation_url directly

Fryguy:


bdunne (commented):
Overall, I like that it's much simpler.  The only down side is that we lose versioning.
Maybe we can inject it?
https://docs.rubocop.org/rubocop/cops_lint.html#lintuselessassignment vs https://docs.rubocop.org/rubocop/1.13/cops_lint.html#lintuselessassignment

Fryguy:
@bdunne Updated.  I don't _love_ this code, but I agree to some extent.  I would like it better if upstream's documentation_url put the correct version in there.  I'm considering coding that upstream, so that's why I put a version check in this code to not insert the version if it already exists.

bdunne (approved):
👍🏻 Thanks

PR #564, opened 2021-06-08
Move to CentOS Stream

Fryguy:
@bdunne I tried updating the bot, but I'm getting errors when building rugged.  I'm surprised I didn't get the same errors with the manageiq core pods.


bdunne:
https://github.com/Fryguy/miq_bot/pull/2

PR #566, opened 2021-08-23
Suggestion for handling Gemfile.lock on miq_bot

agrare:
When I clone this repo and try to run specs I get:

```bash
adam@desktop:~/src/manageiq/miq_bot$ rspec

An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require File.expand_path("../../config/environment", __FILE__)

Gem::MissingSpecVersionError:
  Could not find 'bundler' (1.17.3) required by your /home/grare/adam/src/manageiq/miq_bot/Gemfile.lock.
  To update to the latest version installed on your system, run `bundle update --bundler`.
  To install the missing version, run `gem install bundler:1.17.3`
  Checked in 'GEM_PATH=/home/grare/adam/.gem/ruby/2.7.0:/var/lib/gems/2.7.0:/usr/local/lib/ruby/gems/2.7.0:/usr/lib/ruby/gems/2.7.0:/usr/lib/x86_64-linux-gnu/ruby/gems/2.7.0:/usr/share/rubygems-integration/2.7.0:/usr/share/rubygems-integration/all:/usr/lib/x86_64-linux-gnu/rubygems-integration/2.7.0:/home/grare/adam/.gem' , execute `gem env` for more information
# ./config/boot.rb:4:in `<top (required)>'
# ./config/application.rb:1:in `<top (required)>'
# ./config/environment.rb:2:in `<top (required)>'
# ./spec/spec_helper.rb:3:in `<top (required)>'
No examples found.


Finished in 0.00005 seconds (files took 0.09503 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

If I then `rm Gemfile.lock && bundle update` I pull down an invalid version of `celluloid` which causes failures when running `rspec`:
```bash
adam@desktop:~/src/manageiq/miq_bot$ rspec

An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require File.expand_path("../../config/environment", __FILE__)

LoadError:
  cannot load such file -- celluloid/current
```

Looks like we require `~> 0.17.4`

PR #567, opened 2021-08-24
Update Gemfile.lock

agrare:


NickLaMuro:
@agrare Could we also update `thin` to 1.8.0?  I have run into this issue more times than I can remember, and it gets me every time:

- https://github.com/macournoyer/thin/issues/365
- https://github.com/macournoyer/thin/pull/364

We can do it as a separate PR as well if people prefer.

Gemfile.lock
```diff
@@ -360,25 +354,26 @@ GEM
     sync (0.5.0)
     sysexits (1.2.0)
     temple (0.8.2)
-    thin (1.7.2)
+    thin (1.8.1)
```
agrare, line 357: Absolutely @NickLaMuro :point_up:

NickLaMuro: Oh!  You did a generic `bundle update`, not just for celluloid.  Cool beans.  Thanks!

NickLaMuro (approved):
Looks good to me, but I will let someone else merged like @bdunne or @Fryguy

Gemfile.lock
agrare: Yeah celluloid was the only one which was failing on the latest version so I limited it to the prior version.

spec/models/batch_job_spec.rb
```diff
@@ -85,8 +85,9 @@
 
   describe "#check_complete" do
     it "when destroyed by another checker" do
-      job = described_class.create!.tap(&:destroy)
+      job = described_class.new
```
bdunne, line 88: How is this related to rspec mocks?

agrare: Because a destroyed object is frozen, and rspec 3.10 raises an error about mocking a frozen object (see the error referenced here https://github.com/ManageIQ/miq_bot/pull/566#issuecomment-904082377)

bdunne: So, should we set up the expectation before destroying the record instead?

spec/models/batch_job_spec.rb
```diff
@@ -85,8 +85,9 @@
...
 
+      allow(job).to receive(:reload).and_raise(ActiveRecord::RecordNotFound)
       expect(job).to_not receive(:finalize!)
 
       job.check_complete
```
bdunne, line 93:
```suggestion
      job = described_class.create!

      expect(job).to_not receive(:finalize!)

       job.destroy
      
      job.check_complete
```

spec/models/batch_job_spec.rb
agrare: Sure

Gemfile.lock
kbrock: celluloid was all the rage for a while but died out after sidekick rewrote to not include it.
TBH, kinda surprised to see it.

PR #572, opened 2021-09-30
[MessageBuilder] Include `rubocop-rails` links

NickLaMuro:
Updates COP_URIS so that it also generate correct URLs for cops in `rubocop-rails`, and not just the base gem.

Uses code from the gem to handle generating the URL, and modifies the URL a bit more to properly handle the plugin versions

Links
-----

* https://github.com/rubocop/rubocop/blob/de4f7cdd266999bf2cfbe78bd7fd8a0d4580863d/lib/rubocop/cop/documentation.rb
* https://github.com/rubocop/rubocop/blob/de4f7cdd266999bf2cfbe78bd7fd8a0d4580863d/lib/rubocop/cop/base.rb#L63-L69


Fryguy:
Nice!  I was trying to get something like this in rubocop core, but it was complicated.

PR #579, opened 2022-02-08
Use a matrix for test-repos not jobs

agrare:
There is a limit of 20 on the number of times other workflows can be referenced in a workflow, so rather than 1 job per repo use a matrix for test-repo

### DIFF
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -278,8 +278,8 @@ module GithubService
           "cross-repo" => {
             "uses" => "ManageIQ/manageiq-cross_repo/.github/workflows/manageiq_cross_repo.yaml@master",
             "with" => {
-              "test-repo" => test_repos.inspect,
-              "repos"     => repos.join(",")
+              "test-repos" => test_repos.inspect,
+              "repos"      => repos.join(",")
             }
           }
         }
--- a/lib/github_service/commands/cross_repo_test.rb
+++ b/lib/github_service/commands/cross_repo_test.rb
@@ -274,15 +274,15 @@ module GithubService
         raw_yaml = rugged_repo.blob_at(branch_ref.target.oid, ".github/workflows/ci.yaml").content
         content  = YAML.safe_load(raw_yaml)
 
-        content["jobs"] = test_repos.each_with_object({}) do |test_repo, result|
-          result[test_repo] = {
+        content["jobs"] = {
+          "cross-repo" => {
             "uses" => "ManageIQ/manageiq-cross_repo/.github/workflows/manageiq_cross_repo.yaml@master",
             "with" => {
-              "test-repo" => test_repo,
+              "test-repo" => test_repos.inspect,
               "repos"     => repos.join(",")
             }
           }
-        end
+        }
 
         entry = {}
         entry[:path]  = ".github/workflows/ci.yaml"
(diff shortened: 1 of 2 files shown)


lib/github_service/commands/cross_repo_test.rb
```diff
@@ -274,15 +274,15 @@ def update_github_workflow_yaml_content
...
         content  = YAML.safe_load(raw_yaml)
 
-        content["jobs"] = test_repos.each_with_object({}) do |test_repo, result|
-          result[test_repo] = {
+        content["jobs"] = {
+          "cross-repo" => {
             "uses" => "ManageIQ/manageiq-cross_repo/.github/workflows/manageiq_cross_repo.yaml@master",
             "with" => {
-              "test-repo" => test_repo,
+              "test-repo" => "[#{test_repos.map { |repo| "\"#{repo}\"" }.join(", ")}]",
```
Fryguy, line 281:
```suggestion
              "test-repo" => test_repos.inspect,
```

Ruby array syntax is JSON compliant, so this is a lot simpler.

agrare: Ah thanks much better, felt like there had to be a better way to do that :laughing:

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -274,15 +274,15 @@ def update_github_workflow_yaml_content
...
         content  = YAML.safe_load(raw_yaml)
 
-        content["jobs"] = test_repos.each_with_object({}) do |test_repo, result|
-          result[test_repo] = {
+        content["jobs"] = {
+          "cross-repo" => {
             "uses" => "ManageIQ/manageiq-cross_repo/.github/workflows/manageiq_cross_repo.yaml@master",
             "with" => {
-              "test-repo" => test_repo,
+              "test-repo" => test_repos.inspect,
```
Fryguy, line 281: I wonder if we should change the key to test-repos to make it clear what is incoming

Fryguy (approved):

lib/github_service/commands/cross_repo_test.rb
agrare: Yeah we can do that I'll just need to push another PR to manageiq-cross_repo

agrare: https://github.com/ManageIQ/manageiq-cross_repo/pull/86 https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/583

lib/github_service/commands/cross_repo_test.rb
```diff
@@ -274,15 +274,15 @@ def update_github_workflow_yaml_content
...
-        content["jobs"] = test_repos.each_with_object({}) do |test_repo, result|
-          result[test_repo] = {
+        content["jobs"] = {
+          "cross-repo" => {
             "uses" => "ManageIQ/manageiq-cross_repo/.github/workflows/manageiq_cross_repo.yaml@master",
             "with" => {
-              "test-repo" => test_repo,
-              "repos"     => repos.join(",")
+              "test-repos" => test_repos.inspect,
+              "repos"      => repos.join(",")
```
Fryguy, line 282: Not for this PR, but it might be nice to do this one as `repos.inspect` for consistency with test-repos.  I know that it's not used in the same way, but it's weird to have one be an "Array" and the other not.

PR #586, opened 2022-04-05
Support space deliminated repo names for cross-repo-tests

agrare:
Alternative to fixing https://github.com/ManageIQ/miq_bot/pull/585 is just to support both `" "` and `","` deliminated repo names.  Spaces aren't valid in any of our repo name formats (repo names, PR short format, branch names, urls) so not sure why we explicitly don't allow spaces for separating repo names here.

### DIFF
--- a/spec/lib/github_service/commands/cross_repo_test_spec.rb
+++ b/spec/lib/github_service/commands/cross_repo_test_spec.rb
@@ -148,49 +148,6 @@ RSpec.describe GithubService::Commands::CrossRepoTest do
     end
 
     describe "with invalid repo names" do
-      context "when someone forgets a comma in the test repo list" do
-        let(:command_value) { "manageiq-ui-classic manageiq-api" }
-
-        it "is invalid" do
-          stub_issue_comment(<<~COMMENT)
-            @NickLaMuro 'cross-repo-test(s)' was given invalid repo names and cannot continue
-
-            * `ManageIQ/manageiq-ui-classic manageiq-api`
-          COMMENT
-
-          assert_execute(:valid => false)
-        end
-      end
-
-      context "when someone forgets a comma in the included repo list" do
-        let(:command_value) { "manageiq-ui-classic including manageiq-api#1234 manageiq#1234" }
-
-        it "is invalid" do
-          stub_issue_comment(<<~COMMENT)
-            @NickLaMuro 'cross-repo-test(s)' was given invalid repo names and cannot continue
-
-            * `ManageIQ/manageiq-api#1234 manageiq#1234`
-          COMMENT
-
-          assert_execute(:valid => false)
-        end
-      end
-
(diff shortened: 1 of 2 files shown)


spec/lib/github_service/commands/cross_repo_test_spec.rb
```diff
@@ -148,49 +148,6 @@ def assert_execute(valid: true)
...
-            * `ManageIQ/manageiq-api#1234 manageiq#1234`
-            * `ManageIQ/manageiq-ui-classic manageiq-api`
-          COMMENT
-
-          assert_execute(:valid => false)
-        end
-      end
-
       context "when there's a typo on the including directive" do
         let(:command_value) { "manageiq-ui-classic includes manageiq#1234" }
```
agrare, line 152: This is the only failing test after this change, which I think is just another case of passing an invalid repo name

Fryguy:
I like this approach better.  Does this obviate #585 ?

agrare:
> I like this approach better. Does this obviate https://github.com/ManageIQ/miq_bot/pull/585 ?

Yes, I'll close that and we can continue with this one

spec/lib/github_service/commands/cross_repo_test_spec.rb
Fryguy: Can you add a version of this test that also has space separated and command separated but with the misspelled `inlcuding`? so something like:

```
let(:command_value) { "#{test_repos.join(',')} inlcuding #{repos.join(',')}" }
# and
let(:command_value) { "#{test_repos.join(' ')} inlcuding #{repos.join(' ')}" }
```

agrare: Yes but will end up with the same result which is the command doesn't fail, but the resulting cross-repo-test run will fail trying to pull down `ManageIQ/inlcuding` which we intentionally do not handle here: https://github.com/ManageIQ/miq_bot/blob/master/spec/lib/github_service/commands/cross_repo_test_spec.rb#L329-L336

agrare: Basically before this would be `ManageIQ/manageiq-ui-classic includes manageiq#1234` which we picked up on as invalid because of the spaces, but now it sees `includes` as a repo name and tries to run tests for `ManageIQ/includes`

Not sure if we could catch that here in the command without caching a list of repo names (actually that wouldn't work either because we can pull in other gems from non-ManageIQ orgs)

Fryguy: That's fine...I'll merge this for now, and we can deal with that later.

agrare: Okay let me drop that spec test then so this is green

PR #590, opened 2022-04-28
bundle update

kbrock:
Rails needs to be upgraded

There are some other minor changes
But I can reduct or increase it as need be.


Fryguy:
actions are not set up here yet, so while this is green, it's not actually running the CI.

kbrock:
Well, I'm not sure if we want to use my suggestions, or if you just want to `bundle update` and get a feel for the updates.

Either way, I got a dependabot and wanted to put a place holder in here.

Fryguy:
I'm going to close this, as I needed to to a bundle update as part of the github actions PR #591

Fryguy:
Actually, going to do this the other way around...merge this, then update my Actions PR

PR #603, opened 2022-07-19
Bump Rails to 5.2.8.1

Fryguy:
@bdunne Please review.


config/application.rb
```diff
@@ -24,6 +24,9 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join("lib/github_service/concerns")
     config.eager_load_paths << Rails.root.join("lib")
 
+    # Use yaml_unsafe_load for column serialization to handle Symbols
+    config.active_record.use_yaml_unsafe_load = true
```
kbrock, line 28: Alternatively, we can tell them the classes (e.g.: Symbol, Date, DateTime) that we want to load from serialized form.
It could give us issues with incorrect MiqRequest(Task?) records, but it could nail some things down as well.

Fryguy: Yeah I think I'd like to do that in a follow up... Similarly with ManageIQ core

PR #606, opened 2022-08-03
Warn users about merge commits in PRs

agrare:
A common mistake for developers when updating a branch to include changes from master is to create merge commits like https://github.com/ManageIQ/manageiq-providers-ibm_cloud/commit/da298f3fdf3e9243a89dc2ac62784b16b9f205b6 rather than rebasing their commits on top of the master branch.

We recommend setting `git config pull.rebase true` to avoid this but often times new developers miss this.

### DIFF
--- a/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
+++ b/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
@@ -22,6 +22,7 @@ module CommitMonitorHandlers::CommitRange
 
       new_commits.each do |commit_sha, data|
         check_for_usernames_in(commit_sha, data["message"])
+        check_for_merge_commit(commit_sha, data["merge_commit?"])
       end
 
       @offenses
@@ -59,6 +60,14 @@ module CommitMonitorHandlers::CommitRange
         message = "Username `@#{potential_username}` detected in commit message. Consider removing."
         @offenses << OffenseMessage::Entry.new(:low, message, group)
       end
+    end
+
+    def check_for_merge_commit(commit, merge_commit)
+      return unless merge_commit
+
+      group   = ::Branch.github_commit_uri(fq_repo_name, commit)
+      message = "Merge commit #{commit} detected.  Consider rebasing."
+      @offenses << OffenseMessage::Entry.new(:low, message, group)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
```diff
@@ -22,6 +22,7 @@ def process_commits(new_commits)
 
       new_commits.each do |commit_sha, data|
         check_for_usernames_in(commit_sha, data["message"])
+        check_for_merge_commit(commit_sha, data["merge_commit?"])
```
agrare, line 25: NOTE `merge_commit?` is set in the commit's `details_hash` here: https://github.com/ManageIQ/miq_bot/blob/master/lib/git_service/commit.rb#L78

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
```diff
@@ -60,5 +61,13 @@ def check_for_usernames_in(commit, message)
         @offenses << OffenseMessage::Entry.new(:low, message, group)
       end
     end
+
+    def check_for_merge_commit(commit, merge_commit)
+      return unless merge_commit
+
+      group   = ::Branch.github_commit_uri(fq_repo_name, commit)
+      message = "Merge commit #{commit} detected.  Consider rebasing."
+      @offenses << OffenseMessage::Entry.new(:low, message, group)
```
Fryguy, line 70: Does `:low` translate into ⚠️  or ❕ ?  I think I'd prefer the ⚠️  (probably same goes for `usernames` check)

agrare: I kept the same as the username check, https://github.com/ManageIQ/miq_bot/blob/master/app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb#L60

Fryguy (approved):
LGTM - one minor question

app/workers/commit_monitor_handlers/commit_range/github_pr_commenter/commit_metadata_checker.rb
agrare: Looks like `:low` -> :grey_exclamation:

PR #607, opened 2022-08-04
Build rugged with SSH support

bdunne:
Based on https://github.com/ManageIQ/manageiq/pull/22007


agrare:
@bdunne we also need to bump `rugged` to at least 1.5.0, looks like this is still locked to 1.4.3 :point_right:  https://github.com/ManageIQ/miq_bot/blob/master/Gemfile.lock#L295 (apologies if I've misunderstood this and it is updated during the docker build somehow)

Basically rugged 1.4 broke ssh completely, 1.5 added the `--with-ssh` option

agrare (approved):

Fryguy (approved):

PR #610, opened 2022-08-04
Drop TravisBuildKiller now that we are on GitHub Actions

Fryguy:


lib/github_service/commands/add_label.rb
```diff
@@ -81,18 +83,3 @@ def invalid_label_message(issuer, invalid_labels)
     end
   end
 end
-
-# HACK: Travis / `bundle install --path ...` compat
```
bdunne, line 85: Seems this might still be needed

Fryguy: ugh - ok updated

PR #611, opened 2022-08-04
Replace Sidetiq with Sidecloq

### LINKED ISSUE #282: SideTiq is no longer maintained

Fryguy:
Closes #282

https://github.com/mattyr/sidecloq

@bdunne @agrare Please review.

I need help verifying this because for some reason locally Rugged is having issues with my private keys...it's an unrelated issue, but it prevents me from testing with "real" repos like the sandbox.

Also, switching to Sidecloq should now unlock our Sidekiq version, since we are no longer dependent on celluloid.  I did not do that in this PR, but will do a couple followup PRs to unlock Sidekiq first to 5.2.latest and then to 6

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,8 +26,8 @@ gem 'listen'
 
 # Sidekiq specific gems
 gem 'celluloid', '~> 0.17.4', :require => false
+gem 'sidecloq'
 gem 'sidekiq', '~> 5.2.5', '< 5.2.10'
-gem 'sidetiq', '~> 0.7.2' # TODO: No longer maintained!!
 gem 'sinatra', :require => false
 gem 'slim'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,6 +115,8 @@ GEM
       dry-initializer (~> 3.0)
       dry-schema (~> 1.8, >= 1.8.0)
     erubi (1.10.0)
+    et-orbi (1.2.7)
+      tzinfo
     ethon (0.15.0)
       ffi (>= 1.15.0)
     eventmachine (1.2.7)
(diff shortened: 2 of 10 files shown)


Fryguy:
Note that I intentionally did not include the TravisBuildKiller because I'm going to be killing that anyway in #610

Fryguy:
Updated with a spec to ensure the configured sidecloq queue matches the corresponding class queue as those could go out of sync if you change on and not the other.  It would be nice if sidecloq pulled the queue name from the class if not defined, but alas.

bdunne:
Who is putting things on the queue? Is it just the UI worker? I want to make sure that we don't have everything queued 3 times (UI + 2 queue workers).

Fryguy:
Removed the extra test - turns out you don't need to configure the queues because sidecloq will use the class' queue.

Fryguy:
> Who is putting things on the queue? Is it just the UI worker? I want to make sure that we don't have everything queued 3 times (UI + 2 queue workers).

Sidecloq (much like SideTiq) runs in each sidekiq process that we run, but since those processes share the same redis, a lock is stored in redis and sidecloq runs a leader election where one of the sidekiq processes becomes the leader process.  All of the scheduling happens in that one leader process.  If the leader goes down the other process will become the new leader.

PR #655, opened 2023-11-09
Drop close issue

kbrock:
lets drop the auto close feature for stale PRs

I was not able to run tests locally, so this is more of a conversation starter.

### DIFF
--- a/app/workers/stale_issue_marker.rb
+++ b/app/workers/stale_issue_marker.rb
@@ -29,13 +29,12 @@ class StaleIssueMarker
 
   # Triage logic:
   #
-  # - Stale after 3 month of no activity
-  # - Close after stale and inactive for 3 more months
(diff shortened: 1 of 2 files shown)


app/workers/stale_issue_marker.rb
```diff
@@ -35,32 +35,18 @@ def pinned_labels
   #
   def process_stale_issues
     handle_newly_stale_issues
-    handle_stale_and_unmergable_prs
   end
 
   def handle_newly_stale_issues
-    query  = "is:open archived:false updated:<#{stale_date.strftime('%Y-%m-%d')}"
+    query  = "is:open archived:false -label:stale updated:<#{stale_date.strftime('%Y-%m-%d')}"
```
Fryguy, line 41: I think if we remove the `-label:stale", then this will just comment every 3 months, which is probably what we want.

PR #663, opened 2023-11-21
Upgrade to Rails 6.1 / Ruby 2.7

Fryguy:
This makes a lot more sense if we go commit by commit.

I can pull some of these out as separate PRs to make this easier.

@bdunne @jrafanie I'd like your thoughts on how to move forward.

I tried to get to Ruby 3, but the hard part is actually upgrading faraday, because the specs mess around with the guts of faraday which have changed.


.gitignore
```diff
@@ -39,5 +39,4 @@
 # Ignore local seeds file
 db/seeds.local.rb
 
-# Ignore vagrant configuration/runtime files
-.vagrant/
+/config/master.key
```
Fryguy, line 42: This is something I think we need to put in as a secret in kube that maps to this location, and I guess keep in our vault separately.

Gemfile
```diff
@@ -15,31 +14,27 @@ gem 'jquery-rails'
...
-gem 'awesome_spawn',        '>= 1.4.1'
-gem 'default_value_for',    '>= 3.1.0'
-gem 'haml',                 '~> 5.1',    :require => false # force newer version of haml
-gem 'haml_lint',            '~> 0.35.0', :require => false
-gem 'manageiq-style',                    :require => false
-gem 'more_core_extensions', '~> 4.0.0',  :require => 'more_core_extensions/all'
-gem 'sync',                              :require => false # required by more_core_extensions
-gem 'rugged',                            :require => false
+gem 'minigit', '~> 0.0.4'
+gem 'net-ssh', '~> 7.2.0'
```
Fryguy, line 29: We can probably drop the `~>` here ot use `>=`

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
```
jrafanie, line 5: is it intentional to drop `:password` from this list?  Not sure if this is rails defaults overriding our values.

jrafanie (commented):
A few comments.  Looks good so far.

jrafanie:
Can you update the CI ruby to test against 2.7?

bin/setup
```diff
@@ -1,6 +1,33 @@
...
+
+  puts "\n== Preparing database =="
+  system! 'bin/rails db:prepare'
+
+  puts "\n== Removing old logs and tempfiles =="
+  system! 'bin/rails log:clear tmp:clear'
+
+  puts "\n== Restarting application server =="
+  system! 'bin/rails restart'
+end
```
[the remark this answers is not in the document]
Fryguy: Normalizing to the default bin/setup.

config/initializers/filter_parameter_logging.rb
Fryguy: filter parameters are substring match, so passw includes password.  This is the same original file, so we didn't "add" password originally as it was the default.

bin/setup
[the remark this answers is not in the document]
Fryguy: Actually most of this is the default Rails file - only the part about database.yml.sample was normalizing.

config/environments/development.rb
```diff
@@ -7,23 +7,56 @@
   config.cache_classes = false
 
   # Do not eager load code on boot.
-  config.eager_load = true
+  config.eager_load = false
```
Fryguy: @jrafanie 😭

Gemfile.lock
```diff
@@ -187,18 +186,17 @@ GEM
       net-protocol
     net-ssh (7.2.0)
     nio4r (2.6.0)
-    nokogiri (1.15.5)
-      mini_portile2 (~> 2.8.2)
+    nokogiri (1.15.5-arm64-darwin)
```
Fryguy, line 189: Note to self be sure to bundle lock with the x86_64 arch

app/workers/pull_request_monitor_handlers/merge_target_titler.rb
```diff
@@ -1,31 +1,33 @@
-class PullRequestMonitorHandlers::MergeTargetTitler
-  include Sidekiq::Worker
-  sidekiq_options :queue => :miq_bot
+module PullRequestMonitorHandlers
+  class MergeTargetTitler
```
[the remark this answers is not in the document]
jrafanie: Yes, it might not have been the cause this time but it certainly can cause weird constant lookup strangeness.

bin/setup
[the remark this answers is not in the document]
jrafanie: cool, it wasn't clear if you just took the defaults and changed them after or left them as is.  👍

Fryguy:
> Can you update the CI ruby to test against 2.7?

Done.

Fryguy:
Tests are failing because rails db:prepare doesn't seem to be doing what it's supposed to do on CI - I'll dig in.

jrafanie:
> [Bump tp Ruby 2.7 for GHA](https://github.com/ManageIQ/miq_bot/pull/663/commits/ea4c1ca82154cefa6e3658854c9be53a1be097b4)

noticed this typo in the commit message

jrafanie:
> Tests are failing because rails db:prepare doesn't seem to be doing what it's supposed to do on CI - I'll dig in.

Hmmm, is it possible it's trying to load from schemam.rb?

PR #666, opened 2023-11-22
Update deployment logic for settings changes

bdunne:
- Sync up deployments from the running bot
- Change the settings location to come from a secret now that the defaults are checked in and only secret values are stored in the yml

### DIFF
--- a/templates/bot.yaml
+++ b/templates/bot.yaml
@@ -93,76 +93,86 @@ objects:
 - apiVersion: apps/v1
   kind: Deployment
   metadata:
-    name: postgresql
     labels:
-      app: "bot"
+      app: bot
+    name: postgresql
+    namespace: bot
   spec:
-    strategy:
-      type: Recreate
     replicas: 1
     selector:
       matchLabels:
         name: postgresql
+    strategy:
+      type: Recreate
     template:
       metadata:
-        name: postgresql
         labels:
           name: postgresql
+        name: postgresql
       spec:
-        volumes:
-        - name: pgdb-volume
-          persistentVolumeClaim:
-            claimName: bot-postgres-pvc
-        - name: pg-configs
-          configMap:
-            name: postgresql-configs
-        initContainers:
-        - name: pgsql-data-permission-fix
-          image: busybox
-          command: ["/bin/chmod","-R","777", "/var/lib/pgsql/data"]
-          volumeMounts:
-          - name: pgdb-volume
-            mountPath: "/var/lib/pgsql/data"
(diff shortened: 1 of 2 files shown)


Fryguy (approved):

templates/bot.yaml
```diff
@@ -93,76 +93,89 @@ objects:
...
+          - containerPort: 5432
+            protocol: TCP
+          readinessProbe:
+            failureThreshold: 3
+            initialDelaySeconds: 60
+            periodSeconds: 10
+            successThreshold: 1
+            tcpSocket:
+              port: 5432
+            timeoutSeconds: 1
```
Fryguy, line 147: Original code before the reorder was:

```suggestion
           readinessProbe:
              initialDelaySeconds: 60
              tcpSocket:
                port: 5432
```

templates/bot.yaml
```diff
@@ -93,76 +93,89 @@ objects:
...
-            value: '1000'
+            value: "1000"
           - name: POSTGRESQL_SHARED_BUFFERS
             value: 1GB
+          image: docker.io/manageiq/postgresql:10
+          imagePullPolicy: IfNotPresent
+          name: postgresql
+          ports:
+          - containerPort: 5432
+            protocol: TCP
```
Fryguy, line 139: Technically the protocol was added

templates/bot.yaml
```diff
@@ -93,76 +93,86 @@ objects:
...
+          name: pgsql-data-permission-fix
+          volumeMounts:
+          - mountPath: /var/lib/pgsql/data
+            name: pgdb-volume
+        volumes:
+        - name: pgdb-volume
+          persistentVolumeClaim:
+            claimName: bot-postgres-pvc
+        - configMap:
+            defaultMode: 420
```
Fryguy, line 173: All of these defaultMode look new, so I think we can drop them.

templates/bot.yaml
```diff
@@ -180,44 +190,48 @@ objects:
...
-            claimName: bot-redis-pvc
         containers:
-        - name: redis
-          image: "docker.io/library/redis:5.0"
+        - image: docker.io/library/redis:5.0
+          imagePullPolicy: IfNotPresent
+          name: redis
           ports:
           - containerPort: 6379
+            protocol: TCP
```
Fryguy, line 216: Anoher TCP we can drop

templates/bot.yaml
bdunne: I think we should keep these

PR #679, opened 2024-04-25
Update production settings

Fryguy:
- Add missing repositories and remove old repositories
- Add radjabov labels

### DIFF
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -1,7 +1,11 @@
 core_repos: &core_repos
+- ManageIQ/amazon_ssa_support
+- ManageIQ/container-amazon-smartstate
+- ManageIQ/container-httpd
+- ManageIQ/container-memcached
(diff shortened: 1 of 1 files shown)


bdunne:
Please also add:
```
- ManageIQ/floe
- ManageIQ/manageiq-documentation
```
and remove:
```
- ManageIQ/manageiq-graphql
- ManageIQ/manageiq-v2v
- ManageIQ/manageiq_docs
```

Fryguy:
@bdunne If I recall, the only reason for the core repos list is for the merge target titler, which we only need for those repos where we have the named branches or alternative multiple PR targets.  So we don't need floe, but I do think we need it for repos like container-httpd.  I'll update accordingly.

Fryguy:
@bdunne Updated

agrare (approved):

PR #770, opened 2025-01-22
Upgrade to Rails 7.0 with rake app:update

Fryguy:


Gemfile
```diff
@@ -1,14 +1,11 @@
 source 'https://rubygems.org'
 
-# The Ruby version should match the lowest acceptable version of the application
-raise "Ruby 3.1 required" unless RUBY_VERSION >= "3.1" && RUBY_VERSION < "3.2"
```
Fryguy, line 4: I dropped this because it breaks renovate - it was more about making devs run 3.1 so I removed it for now.

Fryguy:
@jrafanie Please review.

Gemfile
jrafanie: Does `ruby '~> 3.1.0'` do what you want?

Fryguy: > Does `ruby '~> 3.1.0'` do what you want?

Wait you can do the twiddle wack with the ruby command?  I'll give it a try.

jrafanie: yes, https://bundler.io/guides/gemfile_ruby.html

I think we did it this way in manageiq because we wanted a way to warn on untested rubies we wanted to allow but raise on others we explicitly do not want to run on.

Gemfile
```diff
@@ -1,14 +1,11 @@
...
-
 plugin 'bundler-inject'
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
-gem 'rails', '~> 6.1.0', '>= 6.1.7.10'
-gem 'concurrent-ruby', '< 1.3.5' # Temporary pin down as concurrent-ruby 1.3.5 breaks Rails 6.1 & 7.0, and rails-core doesn't
-                                 # plan to ship a new 6.1 or 7.0 to fix it. See https://github.com/rails/rails/pull/54264
+gem 'rails', '~> 7.0.8', '>= 7.0.8.7'
+gem 'concurrent-ruby', '< 1.3.5' # Temporary pin down as concurrent-ruby 1.3.5 breaks Rails 7.0, and rails-core doesn't
```
jrafanie, line 7: :trollface:

Gemfile
Fryguy: No, I just added this line this past week. I tried the ruby command first, but I never thought to use ~>, so it was forcing me to use an exact z stream, and I didn't like that.

jrafanie:
If this is all that's needed to get rails 7.0 upgraded, I'm super jealous.  It's amazing how easy upgrades are if you don't butcher the internals of rails.

Fryguy:
We also only have one controller and view, so it's super tiny 😅

Fryguy:
@jrafanie @bdunne Please review.

jrafanie:
> This pull request is not mergeable. Please rebase and repush.

@Fryguy ^^^. I don't know why I didn't merge this before.  Can you rebase and I'll merge.  😕

PR #825, opened 2025-08-20
[WIP] Upgrade Ruby & Rails to latest versions

bdunne:


spec/spec_helper.rb
```diff
@@ -57,9 +57,10 @@
   config.after do
     Module.clear_all_cache_with_timeout
 
-    # Disable rubocop check because .empty? doesn't exist on a Sidekiq Queue
-    raise "miq_bot queue is not empty" unless Sidekiq::Queue.new("miq_bot").size == 0 # rubocop:disable Style/ZeroLengthPredicate
-    raise "miq_bot_glacial queue is not empty" unless Sidekiq::Queue.new("miq_bot_glacial").size == 0 # rubocop:disable Style/ZeroLengthPredicate
+    # Check that Sidekiq queues are empty before exiting tests
+    require 'sidekiq/api'
+    raise "miq_bot queue is not empty" unless Sidekiq.redis { |c| c.llen("queue:miq_bot") } == 0
+    raise "miq_bot_glacial queue is not empty" unless Sidekiq.redis { |c| c.llen("queue:miq_bot_glacial") } == 0
```
Fryguy, line 63: Very surprising - this is fine for now but I'm surprised there isn't a built-in way to get the queue details from Sidekiq itself.

Gemfile
```diff
@@ -34,19 +33,20 @@ gem 'sidekiq'
...
-gem 'net-ssh', '~> 7.3.0'
+gem 'net-ssh', '~> 7.3'
 
 gem 'awesome_spawn',        '~> 1.6'
 gem 'default_value_for',    '~> 4.0'
-gem 'haml_lint',            '~> 0.51', :require => false
+gem 'haml_lint',            '~> 0.52', :require => false
 gem 'irb'
-gem 'manageiq-style',       '~> 1.5', ">=1.5.6", :require => false
+gem 'manageiq-style',       '~> 1.5.6',  :require => false
```
Fryguy, line 42: This is not the same - think we want to keep the old way.

kbrock: yes. old says 1.6 is ok. new way says only 1.5.7

Gemfile
```diff
@@ -34,19 +33,20 @@ gem 'sidekiq'
...
 gem 'more_core_extensions', '~> 4.4',  :require => 'more_core_extensions/all'
 gem 'rugged',                          :require => false
 
 gem 'faraday'
-gem 'faraday-http-cache', '~> 2.5.0'
-gem 'octokit', '~> 4.25.0', :require => false
+gem 'faraday-http-cache', '~> 2.5'
```
kbrock, line 47: same thing here. do we want 2.5.x or 2.x?

jrafanie:
haha, it's awesome how small upgrades can be if the app isn't tied to internals of rails. ❤️ 🎉 

Does it work?  🤣

Gemfile
kbrock: ok. this looks fine for me.

looks like a slow moving gem and is maintained by someone who looks seasoned and has a lot of gems.

PR #863, opened 2026-01-30
Upgrade Ruby to 3.3

Fryguy:


jrafanie:
Merging so we can see if it goes green after merge.

PR #867, opened 2026-02-02
Upgrade Rails to 7.1

Fryguy:
@jrafanie Please review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,8 +42,9 @@ gem 'manageiq-style',       '~> 1.5', '>=1.5.6', :require => false
 gem 'more_core_extensions', '~> 4.4',  :require => 'more_core_extensions/all'
 gem 'rugged',                          :require => false
 
-gem 'faraday'
+gem 'faraday', '~> 2.14'
 gem 'faraday-http-cache', '~> 2.5.0'
+gem 'faraday-retry'
 gem 'octokit', '~> 4.25.0', :require => false
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -123,6 +123,8 @@ GEM
       faraday (>= 0.8)
     faraday-net_http (3.4.2)
       net-http (~> 0.5)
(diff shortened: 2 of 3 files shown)


config/environments/development.rb
```diff
@@ -6,11 +6,10 @@
...
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.enable_reloading = true
 
-  # NOTE: This is enabled in order to autoload some of the dynamically loaded
-  # classes in development, such as the GithubService::Commands classes
-  config.eager_load = true
+  # Do not eager load code on boot.
+  config.eager_load = false
```
Fryguy: 🎉  - This works on Rails 7.1 with the default!

Fryguy: Just kidding 😭  - Manual testing is showing this not working (wonder if I can find a way to replicate this in tests)

config/environments/development.rb
```diff
@@ -6,10 +6,15 @@
...
-  config.cache_classes = false
+  config.enable_reloading = true
 
   # NOTE: This is enabled in order to autoload some of the dynamically loaded
   # classes in development, such as the GithubService::Commands classes
+  #
+  # Without this change:
+  #   GithubService::CommandDispatcher.registry["add-label"] # => nil
+  # With this change:
+  #   GithubService::CommandDispatcher.registry["add-label"] # => GithubService::Commands::AddLabel
```
Fryguy: My plan for this is to drop the registry altogether and just use Zeitwerk properly. This whole registry is not necessary.

jrafanie (approved):
LGTM, the code loading / changes gives me some concern due to running in threads but if we can a debug bot for a while, we should be able to detect most code loading problems.

config/environments/development.rb
Fryguy: https://github.com/ManageIQ/miq_bot/pull/868

PR #868, opened 2026-02-02
Remove command registry in favor of dynamic class descendants

Fryguy:
@jrafanie Please review.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -8,14 +8,8 @@ Rails.application.configure do
   # since you don't have to restart the web server when you make code changes.
   config.enable_reloading = true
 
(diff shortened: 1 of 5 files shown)


config/environments/development.rb
```diff
@@ -8,9 +8,8 @@
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
 
-  # NOTE: This is enabled in order to autoload some of the dynamically loaded
-  # classes in development, such as the GithubService::Commands classes
-  config.eager_load = true
+  # Do not eager load code on boot.
+  config.eager_load = false
```
Fryguy: This is the main reason I did this - we don't need to eager_load in dev with this change.

PR #869, opened 2026-02-03
Add debugging when builds fail

Fryguy:
@jrafanie Please review. The build in GitHub is detecting source changes, which doesn't make sense to me. I tried the same steps that GitHub tries and I don't see the same thing, which makes me think this is some git configuration thing. Hoping this bit of debug code can help us.


jrafanie (approved):
LGTM, we'll test the build and deploy after merge since local testing couldn't recreate it.

PR #871, opened 2026-02-03
Upgrade to Rails 7.2

Fryguy:
@jrafanie Please review.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -20,7 +20,7 @@ module MiqBot
 
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 7.1
+    config.load_defaults 7.2
 
     # Please, add to the `ignore` list any other `lib` subdirectories that do
     # not contain `.rb` files, or that should not be reloaded or eager loaded.


config/puma.rb
```diff
@@ -1,43 +1,34 @@
...
-# processes).
+# The default is set to 3 threads as it's deemed a decent compromise between
+# throughput and latency for the average Rails application.
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+# Any libraries that use a connection pool or another resource pool should
+# be configured to provide at least as many connections as the number of
+# threads. This includes Active Record's `pool` parameter in `database.yml`.
+threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
+threads threads_count, threads_count
```
jrafanie: We should keep this in mind in case the thread count change affects us.

Fryguy: No, this is fine and we can merge as is. We inherited the default 5 from the old Rails config. The only important number for us is the PostgreSQL pool size, which is currently 25 in the config. The pool size must be at least puma threads + sidekiq threads. We run 2 sidekiq workers with 5 threads each, so that's 3 + 5 + 5 = 13 which is < 25, so we're good.

jrafanie (approved):
LGTM, let's get this in and try it out.

PR #878, opened 2026-02-06
Switch from sidecloq to sidekiq-scheduler

Fryguy:
[sidekiq-scheduler](https://github.com/sidekiq-scheduler/sidekiq-scheduler) is much more well-maintained than [sidecloq](https://github.com/mattyr/sidecloq), so this PR switches the scheduler.  It also supports sidekiq 8, which is what I'd like to upgrade to.

@jrafanie Please review.


config/sidekiq.yml
```diff
@@ -0,0 +1,18 @@
+scheduler:
+  listened_queues_only: true
```
Fryguy: This setting is interesting, and I really like - it basically says "only load schedules for queues that I am watching in this process".  This way even though each sidekiq process is watching different queues, it only tries to schedule for the queue the process is watching.  Everything works fine without this setting, but in the logs it appears like the schedules are loaded twice (because they are), even though they would only fire once. With this setting though it actually only gets loaded where it's expected.

Fryguy:
Not covered by tests. I probably should write some though because I used Bob to convert it and he completely got the YAML file wrong. Totally made up and it didn't work at all. He also didn't know there was a web UI and confidently said, nope there's no UI and deleted the sidecloq one. I only found it because I was looking for the correct YAML syntax.

PR #929, opened 2026-06-02
Change DiffContentChecker to be case insensitive by default

Fryguy:
~~Adds case_sensitive option to control pattern matching behavior. Defaults to true (case-sensitive). Works with both string and regexp types. Extracts regexp building into build_regexp method.~~

This changes the DiffContentChecker to ensure strings are case insensitive by default. We found a few cases where strings were being checked in a case sensitive way, when we expected case insensitive.

@bdunne Please review.

### DIFF
--- a/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
+++ b/spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
@@ -79,4 +79,49 @@ describe CommitMonitorHandlers::CommitRange::GithubPrCommenter::DiffContentCheck
       )
     end
   end
+
+  context "with case-insensitive matching" do
+    context "for string type" do
+      let(:content_1) { "def a(variable)" }
+      let(:content_2) { "PUTS 'hi'" }
+      let(:content_3) { "Puts 'hello'" }
+
+      it "matches case-insensitively" do
+        stub_settings(:diff_content_checker => {"offenses" => {"puts" => {:severity => :error}}})
+        described_class.new.perform(batch_entry.id, branch.id, nil, nil)
+
+        batch_entry.reload
+        expect(batch_entry.result.length).to eq(2)
+        expect(batch_entry.result.first).to have_attributes(
+          :group   => file_path,
+          :locator => 2,
+          :message => "Detected `puts`"
+        )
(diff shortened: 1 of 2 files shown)


GilbertCherrie (approved):

jrafanie:
> Defaults to true (case-sensitive)

I haven't looked.  Are there use cases for case sensitive by default?  I'm not sure what would be more user friendly.

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
```diff
@@ -79,4 +79,106 @@
...
+      it "matches case-insensitively" do
+        stub_settings(:diff_content_checker => {"offenses" => {"puts" => {:severity => :error, :case_sensitive => false}}})
+        described_class.new.perform(batch_entry.id, branch.id, nil, nil)
+
+        batch_entry.reload
+        expect(batch_entry.result.length).to eq(2)
+        expect(batch_entry.result.first).to have_attributes(
+          :group   => file_path,
+          :locator => 2,
+          :message => "Detected `puts`"
```
bdunne, line 98: We probably want `puts` to be case sensitive to avoid false positives

Fryguy: Default is case sensitive like before.

Fryguy:
> > Defaults to true (case-sensitive)
> 
> I haven't looked.  Are there use cases for case sensitive by default?  I'm not sure what would be more user friendly.

We have some things in the config that are code, like puts or binding.pry. those were kept as case sensitive. I guess they could all just be case insensitive and it would probably be the same. 

I'd be ok with switching to only case insensitive.

Fryguy:
@bdunne @jrafanie Updated this to be case insensitive all the time for simplicity

spec/workers/commit_monitor_handlers/commit_range/github_pr_commenter/diff_content_checker_spec.rb
Fryguy: New change is that string settings are case insensitive (regexp settings were already case insensitive)

jrafanie (approved):

PR #936, opened 2026-06-15
Drop legacy code now that octokit has been updated

### LINKED ISSUE #834: Deprecated: Octokit::Client#request_pull_request_review no longer takes a separate :reviewers argument
Noticed this in the logs. We should update this when we bump octokit.

```
Deprecated: Octokit::Client#request_pull_request_review no longer takes a separate :reviewers argument.
Please update your call to pass :reviewers and :team_reviewers as part of the options hash.
```

Fryguy:
This code was added as a temporary placeholder until the upstream PR was merged and octokit updated in this repo. Now that we are on the latest version, this old code can be dropped.

Fixes #834 

@bdunne Please review.

### DIFF
--- a/lib/github_service/commands/remove_reviewer.rb
+++ b/lib/github_service/commands/remove_reviewer.rb
@@ -9,32 +9,14 @@ module GithubService
         user = value.strip.delete('@')
 
         if valid_assignee?(user)
-          if requested_reviewers.include?(user)
-            # FIXME: waiting for merge of https://github.com/octokit/octokit.rb/pull/990
-            begin
-              issue.remove_reviewer(user)
-            rescue NoMethodError
-              # TODO: Remove this exception handling after dependence merge.
-              octokit_request_pull_request_review(issue.fq_repo_name, issue.number, "reviewers" => [user])
-            end
+          if issue.requested_reviewers.include?(user)
+            issue.remove_reviewer(user)
           else
             issue.add_comment("@#{issuer} '#{user}' is not in the list of requested reviewers, ignoring...")
           end
         else
           issue.add_comment("@#{issuer} '#{user}' is an invalid reviewer, ignoring...")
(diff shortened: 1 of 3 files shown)


lib/github_service/commands/remove_reviewer.rb
```diff
@@ -9,33 +9,15 @@ def _execute(issuer:, value:)
...
           else
             issue.add_comment("@#{issuer} '#{user}' is not in the list of requested reviewers, ignoring...")
           end
         else
           issue.add_comment("@#{issuer} '#{user}' is an invalid reviewer, ignoring...")
         end
       end
-
-      # returns an array of user logins who were requested for a pull request review
-      def requested_reviewers
```
Fryguy, line 29: Moved this to the Issue class because it belongs there with the rest of the methods.

lib/github_service/commands/remove_reviewer.rb
```diff
@@ -9,33 +9,15 @@ def _execute(issuer:, value:)
...
-          if requested_reviewers.include?(user)
-            # FIXME: waiting for merge of https://github.com/octokit/octokit.rb/pull/990
-            begin
-              issue.remove_reviewer(user)
-            rescue NoMethodError
-              # TODO: Remove this exception handling after dependence merge.
-              octokit_request_pull_request_review(issue.fq_repo_name, issue.number, "reviewers" => [user])
-            end
+          if issue.requested_reviewers.include?(user)
+            issue.remove_reviewer(user)
```
Fryguy, line 13: Not for this PR, but in a follow up I want to change this to pass all the users at once, because there's no need to support a single-user since the API supports multiple. Technically this is an enhancement.

Down below I kind of started this by passing `Array(users)`, but I kind of needed it for this PR, because the API requires an Array even for a single user.

PR #974, opened 2026-08-26
Recognize WIP after other title tags

### LINKED ISSUE #334: regex in notification monitor for WIP is insufficient
https://github.com/ManageIQ/miq_bot/blob/96d28fab4f605844b02b50911cba2adc114d49f6/lib/github_service/issue.rb#L7 is insufficient as it only looks for WIP at the start.  This can be better handled by reusing the pr_title_tags method: https://github.com/ManageIQ/miq_bot/blob/e81985b7bef0227b963a8372a444178a56a462fe/app/models/branch.rb#L65-L67 , which is what's used by the wip_labeler worker.

See ManageIQ/manageiq#14551 for an example of where it double-wipped a PR.

missusk:
Uses the existing leading title-tag parser when adding or removing the `wip` label. This recognizes `WIP` after tags such as `RFR`, avoids adding a duplicate `[WIP]`, and preserves other leading tags when WIP is removed.

The regression spec fails on `master`: adding the label changes `[RFR] [WIP] Fix title` to `[WIP] [RFR] [WIP] Fix title`, while removing the label leaves the title unchanged.

Fixes #334

Tested:

- `bundle exec rspec spec/lib/github_service/issue_spec.rb spec/models/branch_spec.rb spec/workers/pull_request_monitor_handlers/wip_labeler_spec.rb` (`37 examples, 0 failures`)
- `bundle exec rake` (`365 examples, 0 failures` on Ruby 3.3.10/Linux)

### DIFF
--- a/lib/github_service/issue.rb
+++ b/lib/github_service/issue.rb
@@ -6,7 +6,6 @@ module GithubService
 
     STALE_LABEL       = 'stale'.freeze
     UNMERGEABLE_LABEL = 'unmergeable'.freeze
-    WIP_REGEX         = /^(?:\s*\[wip\])+/i
 
     def assign(user)
       GithubService.update_issue(fq_repo_name, number, "assignee" => user)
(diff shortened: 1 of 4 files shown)


Fryguy:
Nice fix! Thank you @missusk!

lib/github_service/issue.rb
```diff
@@ -118,15 +117,20 @@ def repo_name
...
     end
 
     def unwipify_title
-      if (match = title.match(WIP_REGEX))
-        update(:title => match.post_match.lstrip)
-      end
+      tags = GithubService.title_tags(title)
+      return unless tags.any? { |tag| tag.casecmp?("wip") }
+
+      title_without_tags = title.sub(/^(?:\s*\[[\w-]+\])+/, "").lstrip
```
Fryguy: Minor, but this Regex matches the one in the GitbubService.title_tags method. Can you dedup these regexes - maybe with a constant or a method on the GithubService (i.e. GithubService.strip_title_tags or perhaps GithubService.split_title_tags or something like that)?

missusk: Deduplicated in b1f9099 with `GithubService::TITLE_TAGS_REGEX`.

Fryguy (approved):

Fryguy (approved):

